PR #25, opened 2015-06-08
Updated README to add trailing commas

limedaring:
It's idiomatic Python to have trailing commas, see: https://docs.python.org/2/faq/design.html#why-does-python-allow-commas-at-the-end-of-lists-and-tuples

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ to_address = easypost.Address.create(
   city = "St. Albert",
   state = "AB",
   zip = "t8n2m4",
-  country = "CA"
+  country = "CA",
 )
 from_address = easypost.Address.create(
   name = "Jon Calhoun",
@@ -49,7 +49,7 @@ from_address = easypost.Address.create(
   city = "San Francisco",
   state = "CA",
   zip = "94107",
-  phone = "415-456-7890"
+  phone = "415-456-7890",
 )
 
 # verify address
@@ -65,7 +65,7 @@ if hasattr(verified_from_address, 'message'):
 try:
   parcel = easypost.Parcel.create(
     predefined_package = "InvalidPackageName",
-    weight = 21.2
+    weight = 21.2,
   )
 except easypost.Error as e:
   print e.message
@@ -77,7 +77,7 @@ try:
     length = 10.2,
     width = 7.8,
     height = 4.3,
-    weight = 21.2
+    weight = 21.2,
   )
 except easypost.Error as e:
   raise e
@@ -89,7 +89,7 @@ customs_item = easypost.CustomsItem.create(
   origin_country = "US",
   quantity = 2,
   value = 96.27,
-  weight = 21.1
+  weight = 21.1,
 )
 customs_info = easypost.CustomsInfo.create(
   customs_certify = 1,
@@ -100,7 +100,7 @@ customs_info = easypost.CustomsInfo.create(
   non_delivery_option = "return",
   restriction_type = "none",
   restriction_comments = "",
-  customs_items = [customs_item]
+  customs_items = [customs_item],
 )
 
 # create shipment
@@ -108,7 +108,7 @@ shipment = easypost.Shipment.create(
   to_address = to_address,
   from_address = verified_from_address,
   parcel = parcel,
-  customs_info = customs_info
+  customs_info = customs_info,
 )
 
 # buy postage label with one of the rate objects


att14:
Trailing commas are allowable. PEP8 does not mention this so I'm just going to leave as is.

shazow:
Always having trailing commas makes copying/moving code around a lot more convenient and reduces bugs. Especially since these examples live in the README, ripe for copying/pasting. For example:

```python
foo = [
    1,
    2,
    3
]
```

If you swap the lines that contain 2 and 3, you get a syntax error where you're missing a trailing comma. If you add another line after 3 without diligently noticing the lack of comma, you get another syntax error.

Yes, this isn't specifically addressed in PEP8 but harmless improvements shouldn't be rejected on a technicality.

att14:
Commas at the end of lines can actually lead to bugs, not syntax errors. Say you have the code:

```python
def func(one, two):
  pass

func(
  "foo",
  "bar",
)
```

Now someone refactors:

```python
def func(one=None, two=None):
  pass

arg="bar",

func(
  "foo",
  arg,
)
```

Welp, `arg` is now a 1-tuple and you've got a bug. I see merits to both. I have an aesthetic preference to keep it the way it is.

att14:
Additionally, I've sent a message to the Python docs mailing list to see if there is any merit in updating its copy to highlight this problem.

shazow:
`arg = "bar",` is not the same context—that's a variable definition, a tuple in this case. The original example is function arguments.

```python
foo = [
    1,
    2
    3,
]
```

is a syntax error.

shazow:
Here are some more links:

https://stackoverflow.com/questions/7992559/python-tuple-comma-syntax-rule
https://stackoverflow.com/questions/11597901/why-does-python-allow-a-trailing-comma-in-list
https://coderwall.com/p/4fvcaa/trailing-comma-in-python-data-structures
https://dev.launchpad.net/PythonStyleGuide#Multiline_braces
https://stackoverflow.com/questions/6388187/what-is-the-proper-way-to-format-a-multi-line-dict-in-python

I've never encountered a guide advocating avoiding trailing commas for multi-line lists/dicts/etc in Python. On the other hand, that's very common in JavaScript land where IE refuses to parse trailing commas and thus it didn't make it into the JSON spec. Luckily this is Pythonland. :P

att14:
It works just the same in your case. Let say I want to conditionalize one of the elements of the list.

```python
if some_condition
  element = 2,
else:
  element = None

foo = [
  1,
  2,
  element,
]
```

This is a bug and a confusing one at that. Because the error is going to manifest itself somewhere else in the stack.

att14:
Anyways, I'm not advocating you to not use trailing commas. Be my guest, but it is a matter of personal preference. And my preference is to not use them.

att14:
Lets just say that when you are refactoring anything to do with a comma you have to be careful, no matter if they use a trailing comma or not.

shazow:
```python
if some_condition
  element = 2,
else:
  element = None

foo = [
  1,
  2,
  element,
]
```

... That's not the same thing? Python doesn't do string substitution for variables. Putting commas, or any other character, in random places is not going to be helpful.

> Anyways, I'm not advocating you to not use trailing commas. Be my guest, but it is a matter of personal preference. And my preference is to not use them.

Sure, your codebase. Take care.

att14:
I'm not putting the comma in a random place. Im just copy/pasting the element of the list and assigning it to a variable during the refactor. If I don't realize that there is a trailing comma I've actually changed the type of the element from an integer to a tuple.

I've run into this bug a few times and they are always obnoxious to track down.

shazow:
Copying and pasting a subset of a line into a subset of another line is not the same thing as moving a full line above or below. Your example is equivalent to me taking `foo = 1` and appending the line `abc(1)` into `foo = 1abc(2)`. Of course 1abc doesn't exist, you just concatenated two things that were unrelated so it's going to cause a bug, even without any commas.

On the other hand, much of the time you can move a *full line* around and it'll work great (unless it's scope-specific, like indents). If you don't use trailing commas in multi-line arguments, then you remove that feature and introduce another class of bugs you'll have to deal with with no benefit.

Anyways, if you fully understand this and disagree then that's fine. If not, it would be wise to read the links above (especially the first one Tracy provided)—you and your code will benefit from this style.

But seriously, I acknowledge that this is your codebase and you're free to do what makes you happy. :) I've been writing Python for quite a while so I was hoping to share the benefit of my experience.

att14:
I know who you are :) I've worked with Python for quite a while and to be honest I didn't write any of this code. However, I still stand by no matter if you use trailing commas or not, whenever you refactor anything with commas you must be careful. So I don't think there is a benefit to one way or another.

PR #40, opened 2016-07-17
add to_json method

ThePsyjo:
Add to_json method to get plain json from an EasyPostObject.


victoryftw:
Thanks for adding this. Seems like a nice, simple, useful feature. I'll release a new version containing it shortly.

PR #43, opened 2016-12-03
Add language classifiers to setup.py

dsanders11:
Without a Python3 classifier the tool `caniusepython3` will erroneously flag this package as not supporting Python 3.

You probably want to add more classifiers (see [PayPal-Python-SDK](https://github.com/paypal/PayPal-Python-SDK/blob/master/setup.py) for examples) to help make the package easier to find on PyPI.


Roguelazer:
This is being handled in #49.

PR #44, opened 2016-12-05
first attempt client lib python for reports

codingnutty:


codingnutty:
@victoryftw please let me know what you think of this

easypost/__init__.py
```diff
@@ -902,6 +903,63 @@ def api_keys(self):
...
 
+class Report(AllResource, CreateResource):
+    report_types = { 'shprep': 'shipment', 'plrep': 'payment_log', 'trkrep': 'tracker' }
+
+    @classmethod
+    def create(cls, api_key=None, **params):
+        requestor = Requestor(api_key)
+        url = cls.class_url()
+        wrapped_params = {cls.class_name(): params}
+        report_types = ['shipment', 'payment_log', 'tracker']
```
victoryftw: This array seems to be repeated here and in the `all()` method. Seems like it might be better to pull it out into a class level variable.

easypost/__init__.py
```diff
@@ -77,7 +77,8 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
         'PickupRate': PickupRate,
         'PostageLabel': PostageLabel,
         'CarrierAccount': CarrierAccount,
-        'User': User
+        'User': User,
+        'Report': Report,
```
victoryftw: You may need to add a similar mapping to the `prefixes` below (I can't comment on that line, stats on 84)

victoryftw:
Hey I still don't see any new commits on this PR since my last comment; is everything pushed up as expected or am I missing something?

easypost/__init__.py
```diff
@@ -902,6 +903,63 @@ def api_keys(self):
...
+    report_types = { 'shprep': 'shipment', 'plrep': 'payment_log', 'trkrep': 'tracker' }
+
+    @classmethod
+    def create(cls, api_key=None, **params):
+        requestor = Requestor(api_key)
+        url = cls.class_url()
+        wrapped_params = {cls.class_name(): params}
+        report_types = ['shipment', 'payment_log', 'tracker']
+
+        if str(params['type']) in report_types.values():
```
victoryftw: Arrays do not have a `values()` method; seems like you can just remove the `report_types` array you on line 914

easypost/__init__.py
```diff
@@ -902,6 +903,63 @@ def api_keys(self):
...
+            instance.refresh()
+            return instance
+
+    @classmethod
+    def all(cls, api_key=None, **params):
+        requestor = Requestor(api_key)
+        url = cls.class_url()
+        report_types = ['shipment', 'payment_log', 'tracker']
+
+        if str(params['type']) in report_types:
```
victoryftw: Hmm this one doesn't call `.values()`, but I still think you can remove the redefined `report_types` array.

easypost/__init__.py
```diff
@@ -902,6 +903,63 @@ def api_keys(self):
...
+        try:
+            easypost_id = easypost_id['id']
+        except (KeyError, TypeError):
+            pass
+
+        url = cls.class_url()
+
+        obj_id = params['public_id'].split("_")[0]
+
+        if obj_id in report_types  :
```
victoryftw: There's some weird spacing on this line

tests/report.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import easypost
+from constants import API_KEY as api_key
+
+easypost.api_key = api_key
+
+class ReportTests(unittest.TestCase):
+
+  def test_report_creation(self):
```
victoryftw: This test does not pass; looks to be using Parcel instead of Report for some reason

tests/report.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import easypost
+from constants import API_KEY as api_key
+
+easypost.api_key = api_key
+
+class ReportTests(unittest.TestCase):
```
victoryftw: It'd be great if we could get some tests that test create for the other two types and all/retrieve for all three

easypost/__init__.py
codingnutty: oh damn i added a dictionary with the same name and forgot to get rid of the array.. silly me

easypost/__init__.py
codingnutty: added values.. i started of with an array in the previous commit but added a hash and didnt remove the array

codingnutty:
@victoryftw all the tests now pass and suggested changes have been done. both tests and examples pass..

easypost/__init__.py
victoryftw: These lists are still incomplete: you need to add "ShipmentReport" as class type here and "shprep" to the prefixes list below so that they are properly interpreted by the client as Report objects. Right now it is just parsing Reports retrieved from the system as `easypost.EasyPostObject` not `easypost.Report`.

Same for the other report types, of course

tests/report.py
```diff
@@ -0,0 +1,24 @@
...
+
+  def test_report(self):
+    report = easypost.Report.create(
+        start_date="2012-12-01",
+        end_date="2013-01-01",
+        type="shipment"
+    )
+
```
victoryftw: Since the classing is currently not working as intended, I'd consider adding something like `assert report.__class__ == easypost.Report` after the create and retrieve steps to confirm that it is properly casting to a Report

PR #46, opened 2016-12-20
Use requests session

DorianGray:
Tested in 2.7 and 3.5, seems to work. 3.5 warns about ssl sockets not being closed due to pooling. We really need to rewrite/refactor this(so we can properly close the pool when the "Requestor" is destroyed, but also other reasons).


att14:
@victoryftw can you test and release this?

victoryftw:
LGTM 🚢 ing it

PR #48, opened 2017-02-28
Updates example EasyPost HQ address in the README; adds 'test.sh'; ma…

dlc37:
…kes all tests runnable as __main__.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,11 @@ to_address = easypost.Address.create(
 from_address = easypost.Address.create(
   verify=["delivery"],
   name = "EasyPost",
-  street1 = "118 2nd Street",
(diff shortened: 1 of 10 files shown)


att14:
I agree we should document how to run tests, but I would rather rely on Python's built in unit-testing system rather than introducing a standalone bash script for this. https://docs.python.org/2/library/unittest.html

@victoryftw can you add some documentation and take a look at the rest of this review?

dlc37:
@Roguelazer has made running the tests pythonic in https://github.com/EasyPost/easypost-python/pull/49. There's still the need to make 'if __name__ == "__main__"' consistent -- we do it sometimes, not all the time -- but I'll do that in a separate PR.

PR #50, opened 2017-03-01
modernize tests

Roguelazer:
- get rid of unittest usage (which is inconsistent between py26 and py-not26 anyway)
 - get rid of "classes" that weren't actually classes
 - use conftest.py to configure api_key consistently and remove a bunch of duplicated code
 - make the tests actually work in parallel so that more than one person can run them at a time
 - add python 3.x to the travis config

CC @att14


README.md
```diff
@@ -2,136 +2,154 @@
...
-  )
+    parcel = easypost.Parcel.create(
+        predefined_package = "Parcel",
+        weight = 21.2
+    )
 except easypost.Error as e:
-  print e.message
-  if e.param != None:
-    print 'Specifically an invalid param: ' + e.param
+    print(e.message)
```
dlc37: print(str(e)) works in python3

PR #51, opened 2017-03-01
Dicts iterate differently on python 2 vs. 3. Let's sort.

dlc37:
Inspired by python3 test failures before @Roguelazer modernized them. The order of 'all()' changes.

### DIFF
--- a/tests/test_report.py
+++ b/tests/test_report.py
@@ -11,7 +11,7 @@ def test_shipment_report():
     )
 
     assert report.object == "ShipmentReport"
-    assert report.status == "available"
+    assert report.status in ("available", "new")
     assert report.__class__ == easypost.Report
 
     report2 = easypost.Report.retrieve(report.id)
@@ -32,7 +32,7 @@ def test_payment_log_report():
     )
 
     assert report.object == "PaymentLogReport"
-    assert report.status == "available"
+    assert report.status in ("available", "new")
     assert report.__class__ == easypost.Report
 
     report2 = easypost.Report.retrieve(report.id)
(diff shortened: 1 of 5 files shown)


Roguelazer:
I actually manually added the `message` field to our Exception subclass in #50, since otherwise the APIs were incompatible between Python 2 and Python 3.

I'm fine with this if tests pass, though.

dlc37:
One of the tests failed; I've fixed the test. Not sure if Travis is rerunning the tests automagically...

dlc37:
Tests pass.

tests/test_webhook.py
```diff
@@ -28,7 +28,11 @@ def test_webhooks(per_run_unique):
 
     # Index webhooks
     webhooks = easypost.Webhook.all(url=expected_url)
-    assert webhooks["webhooks"][len(webhooks["webhooks"]) - 1].id == webhook.id
+    for wh in webhooks["webhooks"]:
```
Roguelazer: how about just

`assert any(wh.id == webhook.id for wh in webhooks['webhooks'])`

dlc37: Done, thanks.

tests/test_report.py
```diff
@@ -11,7 +11,7 @@ def test_shipment_report():
     )
 
     assert report.object == "ShipmentReport"
-    assert report.status == "available"
+    assert report.status in ("available", "new")
```
dlc37: Do you think this is sufficient, or will the retrieve() call fail for new reports? I can't reproduce new reports.

Roguelazer: I think it's sufficient. If not, we'll fix it next time it breaks.

Roguelazer:
🛳it

PR #56, opened 2017-04-04
report type (shipment, tracker, payment_log, etc) need to be passed for create/retrieve/all

k-funk:
thoughts on this? It's a bit odd for the type to be a positional param on `create`, but don't want it to be inconsistent with `retrieve`/`all`. If we made them all kwargs, then `retrieve would be odd: `retrieve(type='shipment', id='shprep_...')`/`retrieve('shprep_...', type='shipment')`

Also, these signatures don't match the parent class. Not sure if we care.


easypost/__init__.py
```diff
@@ -950,52 +950,31 @@ def api_keys(self):
 
 class Report(AllResource, CreateResource):
 
-    REPORT_TYPES = {'shprep': 'shipment', 'plrep': 'payment_log', 'trkrep': 'tracker'}
-
     @classmethod
-    def create(cls, api_key=None, **params):
+    def create(cls, type, api_key=None, **params):
```
att14: `type` is a very important function in Python. You cannot use the name.

PR #57, opened 2017-04-24
Support using a custom timeout for requests

jamesbeith:
This change adds support for using a custom timeout. It can be set similar to how `api_key` is set.

```
import easypost
easypost.api_key = '<YOUR API KEY>'
easypost.timeout = 30
```

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -18,9 +18,13 @@ version_info = tuple(int(v) for v in VERSION.split('.'))
 
 # use urlfetch as request_lib on google app engine, otherwise use requests
 request_lib = None
+# use a max timeout equal to that of all customer-facing backend operations
+_max_timeout = 90
 try:
     from google.appengine.api import urlfetch
     request_lib = 'urlfetch'
+    # use the GAE application-wide "deadline" (or its default) if it's less than our existing max timeout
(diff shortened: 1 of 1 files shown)


Roguelazer:
Hi @jamesbeith:

Thanks for the contribution! This is definitely something we should do. Just a couple of thoughts:

(1) We should throw an error if a user attempts to set a timeout value higher than 90 seconds , which is the timeout for all of our customer-facing backend operations.

(2) It's worth noting that this probably won't work exactly as one might expect on GAE yet; App Engine has a default application-wide "deadline", and we set a 55-second timeout for our requests on line 358. It might be worth changing line 358 to `args['deadline'] = timeout` and adding something at module scope like:

```
try:
    from google.appengine.api import urlfetch
    if timeout > google.appengine.api.urlfetch.get_default_fetch_deadline():
        raise ValueError('You must set the GAE global deadline higher than the EasyPost timeout value')
except ImportError:
    pass
```

Otherwise there's the confusion that somebody could set a timeout here and then have it not be honored.

What do you think?

jamesbeith:
Hey @Roguelazer. I've made your suggested changes. Let me know if you have any further comments or suggestions.

jamesbeith:
Hey @att14 and @Roguelazer, is there anything I can do to help move this along? Cheers

jamesbeith:
Much appreciated

PR #58, opened 2017-04-26
Make no rates found error optional

fengsi:
It is unnecessary to always raise error if no rates could be found.  Raising error made it inconvenient to use as you were required to have a redundant wrapper around the possible error.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -753,13 +753,13 @@ class Shipment(AllResource, CreateResource):
             carriers = carriers.split(',')
         except AttributeError:
             pass
-        carriers = [c.lower() for c in carriers]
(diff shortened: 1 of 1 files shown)


Roguelazer:
It's not particularly Pythonic to make errors optional like this. Can you tell us a little more about your use cases?

As for your other changes (changing the services / carriers to `set`s), I don't particularly object to it; I don't think our APIs ever return duplicates and I'm positive that any performance difference for iterables this short is in the margin of error, but _shrug_. As a tip, you should [nearly always] construct sets from generators `set(i.lower() for i in iterable)` instead of from arrays `set([i.lower() for i in iterable])`. No sense making the GC do more work than necessary.

PR #64, opened 2017-06-27
added standalone rating endpoint

algreen11:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ### NEXT (unreleased)
 
+### 4.0.0 2017-06-26
```
k-funk: Are there breaking changes? If not, I'd say that `3.7` would be a fine version #.

easypost/__init__.py
```diff
@@ -291,21 +294,26 @@ def request_raw(self, method, url, params=None, apiKeyRequired=True):
...
             ua['openssl_version'] = ssl.OPENSSL_VERSION
-
-        headers = {
-            'X-Client-User-Agent': json.dumps(ua),
-            'User-Agent': USER_AGENT,
-            'Authorization': 'Bearer %s' % my_api_key,
-            'Content-type': 'application/x-www-form-urlencoded'
-        }
-
+        if isRating:
```
k-funk: no need to recreate the dict a second time. could just create the headers, and then replace the key.
```
headers = {...}
if isRating:
    headers['Content-Type'] = 'application/json'
```

easypost/__init__.py
```diff
@@ -291,21 +294,26 @@ def request_raw(self, method, url, params=None, apiKeyRequired=True):
...
+                'User-Agent': USER_AGENT,
+                'Authorization': 'Bearer %s' % my_api_key,
+                'Content-type': 'application/json' 
+            }
+        else:
+            headers = {
+                'X-Client-User-Agent': json.dumps(ua),
+                'User-Agent': USER_AGENT,
+                'Authorization': 'Bearer %s' % my_api_key,
+                'Content-type': 'application/x-www-form-urlencoded'
```
k-funk: Why was "type" (Content-type) not capitalized?
Can you change it and test that it still works correctly?

easypost/__init__.py
```diff
@@ -321,27 +329,40 @@ def interpret_response(self, http_body, http_status):
...
-            data = self.encode(params)
+            if not isRating:
+                data = self.encode(params)
+            else:
+                data = params
         else:
             raise Error("Bug discovered: invalid request method: %s. "
                         "Please report to contact@easypost.com." % method)
-
+
```
k-funk: whitespace

easypost/__init__.py
```diff
@@ -321,27 +329,40 @@ def interpret_response(self, http_body, http_status):
...
-    def requests_request(self, method, abs_url, headers, params):
+    def requests_request(self, method, abs_url, headers, params, isRating):
         method = method.lower()
         if method == 'get' or method == 'delete':
             if params:
                 abs_url = self.build_url(abs_url, params)
             data = None
         elif method == 'post' or method == 'put':
-            data = self.encode(params)
+            if not isRating:
```
k-funk: This seems pretty hacky. It seems like the check that you want here is if `headers['Content-Type'] == 'application/json'` (beware the capital T as mentioned above) vs passing `isRating` into this method.

easypost/__init__.py
```diff
@@ -997,3 +1018,12 @@ def update(self, **params):
         response, api_key = requestor.request('put', url, params)
         self.refresh_from(response, api_key)
         return self
+
+class Rating(CreateResource):
+    @classmethod
+    def create(cls, api_key=None,**params):
```
k-funk: space between 2nd and 3rd param

examples/rating.py
```diff
@@ -0,0 +1,35 @@
+import easypost
+
+easypost.api_key = 'EASYPOSTAPIRATINGACTIVATEDKEY'
+
+rating = easypost.Rating.create(
+  carrier_accounts=["CARRIERACCOUNTID1", "CARRIERACCOUNTID2"],
+  to_address={
+    "name": 'Dr. Steve Brule',
```
k-funk: i know that the library is currently inconsistent about single vs double quotes. but i'm changing it to single quotes in my upcoming release. could you change all double quotes to single quotes here so that i can just copy this example in?

easypost/__init__.py
```diff
@@ -321,27 +329,40 @@ def interpret_response(self, http_body, http_status):
...
+                    data=data,
+                    timeout=timeout,
+                    verify=True,
+                )
+            else:
+                result = requests_session.request(
+                    method,
+                    abs_url,
+                    headers=headers,
+                    data=json.dumps(data),
```
k-funk: this could also be done above on line 339 by just reassigning `data=json.dumps(data)`. it's simpler to keep all this edge-case logic in one place

easypost/__init__.py
algreen11: Weird, I guess it isn't case sensitive

easypost/__init__.py
```diff
@@ -328,12 +330,14 @@ def requests_request(self, method, abs_url, headers, params):
...
         elif method == 'post' or method == 'put':
-            data = self.encode(params)
+            if headers['Content-Type'] != 'application/json':
+                data = self.encode(params)
         else:
             raise Error("Bug discovered: invalid request method: %s. "
                         "Please report to contact@easypost.com." % method)
-
         try:
+            if headers['Content-Type'] == 'application/json':
```
k-funk: Could probably move this up and merge with lines 333/334 so that you have an if/else.
```
if application/json:
   do these abnormal things
else:
   do the normal things
```
If the json.dumps truly needs to be try/except-wrapped (not sure if it does), you'll probably want to catch whatever exception that it throws so that debugging is easier.

easypost/__init__.py
```diff
@@ -291,25 +294,24 @@ def request_raw(self, method, url, params=None, apiKeyRequired=True):
 
         if hasattr(ssl, 'OPENSSL_VERSION'):
             ua['openssl_version'] = ssl.OPENSSL_VERSION
-
         headers = {
             'X-Client-User-Agent': json.dumps(ua),
             'User-Agent': USER_AGENT,
             'Authorization': 'Bearer %s' % my_api_key,
-            'Content-type': 'application/x-www-form-urlencoded'
+            'Content-Type': 'application/x-www-form-urlencoded'
```
k-funk: extra whitespace

VERSION
```diff
@@ -1 +1 @@
-3.6.2
+3.7.0
```
k-funk: only nitpicking because there's other changes to be done as well, but:
files should have newlines at the end. I can't remember the reason why... but that's just what we do, and what the github symbol at the end of this line means :P

easypost/__init__.py
```diff
@@ -272,8 +273,10 @@ def request_raw(self, method, url, params=None, apiKeyRequired=True):
                 'at contact@easypost.com for assistance.')
 
         abs_url = self.api_url(url)
+        if (url == "/rating/v1/rates"):
```
k-funk: To me, I would have not use the `isRating` var on line 264/278, and would have done this `abs_url` re-assignment down on line 303 where the header is being modified, so that this is all in one place. However, the url assignment would now be happening far away from the original `abs_url` assignment, which I don't like. @DorianGray thoughts?

(don't make this change unless someone else weighs in)

VERSION
algreen11: good to know the convention, I had no idea

k-funk: rule of thumb:
major 3.x.x: incompatible with previous versions
minor x.3.x: new features
maintenance x.x.3: bugfixes

easypost/__init__.py
```diff
@@ -997,3 +1001,12 @@ def update(self, **params):
         response, api_key = requestor.request('put', url, params)
         self.refresh_from(response, api_key)
         return self
+
```
Roguelazer: there should be two blank lines between top-level definitions (PEP8 E302)

easypost/__init__.py
```diff
@@ -997,3 +1001,12 @@ def update(self, **params):
...
+class Rating(CreateResource):
+    @classmethod
+    def create(cls, api_key=None, **params):
+        requestor = Requestor(api_key)
+        url = "/rating/v1/rates"
+        response, api_key = requestor.request('post', url, params)
+        return convert_to_easypost_object(response, api_key)
+
```
Roguelazer: there should not be a blank line at the end of the file (just a newline) (PEP8 W391)

easypost/__init__.py
```diff
@@ -261,6 +261,7 @@ def request(self, method, url, params=None, apiKeyRequired=True):
         return response, my_api_key
 
     def request_raw(self, method, url, params=None, apiKeyRequired=True):
+        isRating = False
```
Roguelazer: PEP8 prefers snake_case over camelCase

examples/rating.py
```diff
@@ -0,0 +1,35 @@
+import easypost
+
+easypost.api_key = 'EASYPOSTAPIRATINGACTIVATEDKEY'
+
+rating = easypost.Rating.create(
+  carrier_accounts=['CARRIERACCOUNTID1', 'CARRIERACCOUNTID2'],
```
Roguelazer: indents should always be 4 characters in python source code

examples/rating.py
```diff
@@ -0,0 +1,35 @@
...
+    "email": 'support@easypost.com'
+  },
+  parcels=[{
+    "length": 20.2,
+    "width": 10.9,
+    "height": 5,
+    "weight": 65.9
+  }]
+)
+print(rating)
```
Roguelazer: all files should end with a newline on *nix systems (that's what the red cancel icon here means)

easypost/__init__.py
Roguelazer: this might be more descriptive as "`is_json_only_endpoint`" or somesuch

VERSION
Roguelazer: I would actually recommend separating the version bump into a different commit; typical practice is to land features onto master and then after master passes tests, create an additional commit on top of master that bumps the version (and changes the `CHANGELOG` from "NEXT" to the actual version number).

easypost/__init__.py
Roguelazer: (this is resolved but github isn't smart enough to resolve the comment)

easypost/__init__.py
DorianGray: I agree, it's messy as it is now. Also, comparing/using random strings in line is 99.9% of the time bad form. I would convert them to constants.

examples/rating.py
```diff
@@ -0,0 +1,35 @@
+import easypost
+
+easypost.api_key = 'EASYPOSTAPIRATINGACTIVATEDKEY'
```
DorianGray: `EASYPOSTA PIRATING ACTIVATED KEY`

Justintime50:
Standalone rating is no longer supported. Thank you for your efforts on this PR!

PR #66, opened 2017-09-22
Tuple of carrier choices

cranberyxl:
Add tuple of carrier choices from https://www.easypost.com/docs/api#carrier-tracking-strings that can be used programmatically to display or select carriers on clients.


mwaldt:
Since we build support for new carriers regularly it is recommended that you use our [Retrieve available Carrier Types endpoint](https://www.easypost.com/docs/api#retrieve-available-carrier-types) instead of hardcoding this into the repository.

Issue #67, opened 2018-01-18
CarrierAccount.types api_key not handled correctly


eprussell:
Seems like when there's a field `api_key` in a carrier type this isn't behaving as intended with this client library. 

Example below is Australia Post but this appears to happen for any carrier with api_key among the fields (e.g. Parcel, Fastway, etc.)

result with cURL:
```
curl -X GET https://api.easypost.com/v2/carrier_types   -u PRODUCTION_API_KEY:
```
```
{
 "object": "CarrierType",
 "type": "AustraliaPostAccount",
 "readable": "Australia Post",
 "logo": null,
 "fields": {
  "credentials": {
   "account_number": {
    "visibility": "visible",
    "label": "Australia Post Account Number"
   },
   "api_key": {
    "visibility": "masked",
    "label": "Australia Post API Key"
   },
   "api_secret": {
    "visibility": "password",
    "label": "Australia Post Secret Key"
   },
   "print_as_you_go": {
    "visibility": "checkbox",
    "label": "Print as you go"
   }
  }
 }
},
```
result with python client library:
```
import easypost
easypost.api_key = 'api_key_here'
carrier_types = easypost.CarrierAccount.types()
print carrier_types
```
```
<EasyPostObject CarrierType at 0x11007abd0> JSON: {
  "fields": {
    "credentials": {
      "account_number": {
        "label": "Australia Post Account Number",
        "visibility": "visible"
      },
      "api_secret": {
        "label": "Australia Post Secret Key",
        "visibility": "password"
      },
      "print_as_you_go": {
        "label": "Print as you go",
        "visibility": "checkbox"
      }
    }
  },
  "logo": null,
  "object": "CarrierType",
  "readable": "Australia Post",
  "type": "AustraliaPostAccount"
},
```

Note in the Python case the absence of `api_key`

Roguelazer:
that seems entirely plausible

Roguelazer:
This is fixed in version 3.6.3, now available [in pypi](https://pypi.python.org/pypi/easypost).

PR #68, opened 2018-02-06
Fixed api_key bug.

jaydensmith:
#67 fixes api_key credentials issue.

### DIFF
--- a/tests/test_batch.py
+++ b/tests/test_batch.py
@@ -74,8 +74,8 @@ def test_batch_create_and_buy():
     assert batch.shipments[0].buyer_address.street1 == '388 Townsend St'
 
     # Assert on fees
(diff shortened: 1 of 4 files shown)


tests/test_batch.py
```diff
@@ -74,8 +74,8 @@ def test_batch_create_and_buy():
     assert batch.shipments[0].buyer_address.street1 == '388 Townsend St'
 
     # Assert on fees
-    assert batch.shipments[0].fees[0].amount == '0.03000'
-    assert batch.shipments[0].fees[1].amount == '6.52000'
+    assert batch.shipments[0].fees[0].amount == '0.01000'
```
Roguelazer: my, it has been a while since we touched this client library

PR #71, opened 2018-04-09
Fix for sending delivery and zip4 verification in the same request

abramclark:
Old code constructed broken query params


abramclark:
@att14 how's this?

easypost/__init__.py
```diff
@@ -621,19 +621,15 @@ def delete(self, **params):
 class Address(AllResource, CreateResource):
 
     @classmethod
-    def create(cls, api_key=None, verify=None, verify_strict=None, **params):
+    def create(cls, api_key=None, verify=[], verify_strict=[], **params):
```
att14: You can't use mutable types as kwargs.

abramclark: I suppose this could be called repeatedly. Uhg I'm not a fan of that bit of Python behavior

PR #73, opened 2018-06-30
Support Python 3.7

dsanders11:


dsanders11:
Looks like I jumped the gun here and Travis CI doesn't support Python 3.7 yet (but probably will in the quite near future).

Roguelazer:
I'm going to close this in favor of #86 which seems to be the "preferred" way to do this upstream

PR #83, opened 2019-07-09
- fixed lowest_rate() selector to handle comma separated string values

catokx:


fthobe:
Hi @mwaldt , my colleague @catokx made a grade contribution to your library. Could you take a look as the python library is currently (since roughly 12 hours) broken as DHL changed the format for price quotes.

mwaldt:
Hi,

I pushed an internal fix to the carrier on our end this morning. In the
future I would recommend reaching out to support@easypost.com with
inquiries such as these for more expedited results. I cc'd our support line
here as well for your convenience.

Cheers,

fthobe:
@mwaldt it didn’t work for us. The fix @catokx pushed fixes variations in the format of the price, wouldn’t it make sense to merge it or does it break anything (values >999.99)?

fthobe:
@mwaldt we did speak to customer support, they advised us to open an issue in GitHub!

Justintime50:
@fthobe @catokx has this been an issue since? I believe this has been patched since initially being submitted and wanted to know if indeed it had or not.

PR #84, opened 2019-07-10
fix install_requires typo that prevented 'six' from being installed

roehnan:
The 'setup.py' didn't separate the install_requires list arguments with a comma. I believe that is what was preventing the 'six' library from being installed by pip


Roguelazer:
whoops, you're definitely right

let me figure out why travis is sad.

Roguelazer:
okay if you rebase on master tests will pass now

PR #86, opened 2019-07-10
attempt to add python 3.7 and pypy3.5 to the build matrix

Roguelazer:
replaces #73


Roguelazer:
bah rebasing on master

PR #88, opened 2019-07-10
populate some more setup fields to get a prettier PyPI page

Roguelazer:


setup.py
```diff
@@ -24,11 +35,16 @@
     package_data={'easypost': ['../VERSION']},
     install_requires=install_requires,
     test_suite='test',
+    long_description=long_description,
+    long_description_content_type='text/markdown',
+    project_urls={
+        'CI': 'https://travis-ci.org/EasyPost/easypost-python',
+    },
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
```
victoryftw: the readme still says `Python 2.6, 2.7, or 3.3+ (or corresponding PyPy versions).` but this is >=2.7. Should we update the readme?

Roguelazer: yeah, this hasn't worked on Python 2.6 in a couple of years. I'll update the README.

PR #91, opened 2020-03-04
Fixing the create reports params wrapping issue

Justintime50:
We've had multiple people test and confirm that the Python client library wraps params for reports incorrectly which leads to users not being able to create reports via the client library and have them show up on their Dashboard. With this fix in place, reports will create properly with the correct parameter wrapping and they show up on the Dashboard.

Let me know if I need to do a version bump etc for this or if that's something handled by someone else.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -961,9 +961,7 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         requestor = Requestor(api_key)
         url = "%s/%s" % (cls.class_url(), params['type'])
-        wrapped_params = {cls.class_name(): params}
-
-        response, api_key = requestor.request('post', url, wrapped_params, False)
+        response, api_key = requestor.request('post', url, params, False)
         return convert_to_easypost_object(response, api_key)
 
     @classmethod
@@ -972,7 +970,6 @@ class Report(AllResource, CreateResource):
         url = "%s/%s" % (cls.class_url(), params['type'])
         response, api_key = requestor.request('get', url, params)
         return convert_to_easypost_object(response, api_key)
-
 
 class Blob(AllResource, CreateResource):
     @classmethod
@@ -986,7 +983,6 @@ class Blob(AllResource, CreateResource):
         url = "%s/%s" % (cls.class_url(), easypost_id)
         response, api_key = requestor.request('get', url)
         return response["signed_url"]
-
 
 class Webhook(AllResource, CreateResource, DeleteResource):
     def update(self, **params):


easypost/__init__.py
```diff
@@ -961,11 +961,10 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         requestor = Requestor(api_key)
         url = "%s/%s" % (cls.class_url(), params['type'])
-        wrapped_params = {cls.class_name(): params}
-
-        response, api_key = requestor.request('post', url, wrapped_params, False)
+        response, api_key = requestor.request('post', url, params)
         return convert_to_easypost_object(response, api_key)
 
+
```
jontsai: methods within a class should only have 1 newline in between, per PEP8 (https://www.python.org/dev/peps/pep-0008/)

easypost/__init__.py
```diff
@@ -961,11 +961,10 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         requestor = Requestor(api_key)
```
jontsai: we should add a docstring to this class, if not function

easypost/__init__.py
```diff
@@ -961,11 +961,10 @@ class Report(AllResource, CreateResource):
...
         url = "%s/%s" % (cls.class_url(), params['type'])
-        wrapped_params = {cls.class_name(): params}
-
-        response, api_key = requestor.request('post', url, wrapped_params, False)
+        response, api_key = requestor.request('post', url, params)
```
jontsai: When named-default parameters are present, it is a best practice to invoke the function with named arguments as well, if they are different from the default values.

The signature of the request method, per https://github.com/EasyPost/easypost-python/pull/91/files#diff-311814b232626ffa6fd7feb3ee0aa9a4L256, is:

`def request(self, method, url, params=None, apiKeyRequired=True):`

Would recommend invoking as:

`response, api_key = requestor.request('post', url, params=params)`

This changes from `apiKeyRequired` from `False` to True`

It's not clear why "unwrapping" the params is the right fix

jontsai: Also, CI is failing, so we should:

1. Whenever adding code/bug fix, add an accompanying test
2. Get CI to pass

roehnan:
The issue is that this call incorrectly wraps the parameters within the class name.

The parameters look like:

`{
    "report": {
        "end_date": "2019-8-31",
        "start_date": "2019-08-01",
        "type": "payment_log"
    },
    "format": "json",
    "controller": "reports",
    "action": "create",
    "type": "payment_log"
}`

(underneath 'report')

when they should look like:

`{
    "type": "payment_log",
    "start_date": "2019-08-01",
    "end_date": "2019-08-31",
    "format": "json",
    "controller": "reports",
    "action": "create"
}`

roehnan:
The EasyPost API doesn't honor the start_date and end_date parameters unless they are at the outer level.

roehnan:
Trackers (for example) do not wrap parameters:

image (1)

But Users do:

image (2)

The logic for creating Reports needs to follow the model used by Trackers and not Users.

jontsai:
Thank you @roehnan and @Justintime50 for the additional info and context. The code for this PR looks good now. However, in order to approve and merge this in, we should get the CI passing.

Could we get the CI passing in another PR?

Justintime50:
Thanks @jontsai,

I’m actually working on another PR right now to fix that. It’s odd though as some tests pass in some environments and others do not. I’ll push up a PR soon to fix CI and rebase this PR off those fixes to ensure they pass too.

easypost/__init__.py
Justintime50: I too thought this was strange but tried to keep it similar to what was already there. I've removed the extra lines from the Report class.

easypost/__init__.py
Justintime50: I agree; however it would be the only docstring in the library. Could be a separate PR?

easypost/__init__.py
Justintime50: CI should hopefully pass if I rebased correctly. I've added back in the missing `False`. Kept params as is since it matches the rest of the library where params are also used.

Justintime50:
@jontsai this should now be ready for review. Thanks for the feedback. CI passed!

Closes #92

jontsai:
@Justintime50 this is now merged, so go ahead and create a PR to update version

PR #93, opened 2020-03-06
Fix unit tests

Justintime50:
It's been some time since unit tests were run and some items have changed making a few fail. The following changes are fixes to bring unit tests into the current state of the EasyPost API and carrier requirements so unit tests will pass once again.

### DIFF
--- a/tests/test_pickup.py
+++ b/tests/test_pickup.py
@@ -2,14 +2,11 @@
 
 import time
 import datetime
-
 import easypost
 import pytest
 import pytz
 
-
 ONE_DAY = datetime.timedelta(days=1)
-
 
 @pytest.fixture
 def noon_on_next_monday():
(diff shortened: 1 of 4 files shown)


tests/test_pickup.py
```diff
@@ -2,34 +2,32 @@
...
     return datetime.datetime.combine(next_monday, noon_est)
 
-
 def test_pickup_batch(noon_on_next_monday):
     # Create a Batch containing multiple Shipments. Then we try to buy a Pickup and assert if it was bought.
 
     pickup_address = easypost.Address.create(
         verify=['delivery'],
-        name='Jarrett Streebin',
+        name='TAKASHI KOVACS',
```
k-funk, line 23: new sleeve, new name? 😆

Justintime50: Haha I have all sorts of random names floating on all my scripts, this one just happened to be the latest!

k-funk: I have been littering repos with Roland Deschain lately https://en.wikipedia.org/wiki/Roland_Deschain

Justintime50: Very nice haha. My latest is Jake Peralta.

PR #96, opened 2020-03-19
Closes #95 - Closes the VERSION file

### LINKED ISSUE #95: Unclosed file in "version.py"
Unclosed file in `version.py`:
```
/venv3/lib/python3.6/site-packages/easypost/version.py:4: ResourceWarning: unclosed file <_io.BufferedReader name='/venv3/lib/python3.6/site-packages/easypost/../VERSION'>
  VERSION = pkg_resources.resource_stream('easypost', '../VERSION').read().decode('utf-8').strip()
```
To reproduce add `PYTHONWARNINGS=always` to environment variables.

Justintime50:
An issue was reported about how the VERSION file stays open and is not closed. For best practice sake and to correct the warning, I put a fix in place that closes the file after we access the contents and assign to a variable. Closes #95 .

### DIFF
--- a/easypost/version.py
+++ b/easypost/version.py
@@ -1,7 +1,8 @@
 import pkg_resources
 
-
-VERSION = pkg_resources.resource_stream('easypost', '../VERSION').read().decode('utf-8').strip()
+VERSION_FILE = pkg_resources.resource_stream('easypost', '../VERSION')
+VERSION = VERSION_FILE.read().decode('utf-8').strip()
+VERSION_FILE.close()
 
 if '-' in VERSION:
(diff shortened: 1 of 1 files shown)


Roguelazer:
There's no reason to make this a global symbol; something like

```python
with contextlib.closing(pkg_resources.resource_stream('easypost', '../VERSION')) as _version_file:
    VERSION = _version_file.read()....
```

should be sufficient and more idiomatic

Justintime50:
If we instead change how we handle the version number as per #98 then we'd need to address this differently anyway huh?

PR #104, opened 2020-05-05
stop doing tricks to populate version

### LINKED ISSUE #95: Unclosed file in "version.py"

### LINKED ISSUE #98: DeprecationWarning: Use of .. or absolute paths will raise exceptions...
When `PYTHONWARNINGS=always` is set, I get the following from the pkg_resources library:

```
/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py:1151: DeprecationWarning: 
Use of .. or absolute path in a resource path is not allowed and will raise exceptions in a future 
release.
  self, resource_name
```

We'll need to adjust the relative path to the `VERSION` file in the near future to avoid throwing exceptions on each call. The offending code is in `version.py`:

```python
VERSION_FILE = pkg_resources.resource_stream('easypost', '../VERSION')
```

Roguelazer:
Closes #95 
Closes #98 
Obviates #96


easypost/version.py
```diff
@@ -1,7 +1,7 @@
 import pkg_resources
 
 
-VERSION = pkg_resources.resource_stream('easypost', '../VERSION').read().decode('utf-8').strip()
+VERSION = '4.0.1'
```
jasonbot, line 4: yay

Issue #101, opened 2020-04-23
Undocumented Easypost APIs


sameerkumar18:
What's a good way to extend Easypost python wrapper to support undocumented APIs like `/v2/fedex_registrations` and `v2/carrier_accounts/register` ?
We're aware of calling the endpoints directly with packages like requests, but for maintaining our code design, is it possible to extend the wrapper to support new API resources?

Thanks!

Justintime50:
Hey @sameerkumar18, this is a great request! A few of our other libraries have PR's in progress to add this functionality. We welcome PR's if you'd like to contribute! These endpoints are certainly something to consider.

Until the functionality is added to the client library, you're correct - requests would be a great alternative.

sameerkumar18:
Sure @Justintime50 
I'll give it a shot and update in a few days :)

sameerkumar18:
@Justintime50 Can you take a look at #109 
Please suggest some improvements, if any. Also can you please help me with the failing tests?
Thanks! :D

LoganSimonsen:
Just a note, FedEx CSP registration should no be possible via `v2/carrier_accounts/register` eliminating the need to deal with `/v2/fedex_registrations`

Justintime50:
See #109 for additional details on why this was closed.

PR #109, opened 2020-05-17
Support for Undocumented APIs

sameerkumar18:
List of changes:
- (#101 ) Added support for calling an un-mapped URL via wrapper itself. Custom resource endpoint & request params/body
- Other minor PEP8 & Code cleanups

To Do:
- Adding examples for calling un-mapped APIs


Justintime50:
My apologies, I may not have been clear enough on the direction we'd want to take this.

We have other open PR's that more closely follow the direction we'd want with registering FedEx accounts:
* https://github.com/EasyPost/easypost-java/pull/50/files
* https://github.com/EasyPost/easypost-csharp/pull/131/files

With that being said, we may want to take a more holistic approach to allowing carrier account registration for carriers that have a custom workflow (FedEx, UPS, etc).

As to your PEP8 and code cleanup, most of that looked ok! Feel free to open up a new separate PR that addresses only that (code cleanup etc).

I don't believe we'll want to be adding an `Undocumented` class at this time, we'd rather focus on adding first-class support for each feature of our API one-by-one. I know that registering FedEx accounts is a big feature many want, we'll look into this item soon.

PR #114, opened 2020-08-10
Bump version - 4.2.0

Justintime50:
Bumping the version for all the changes made two weeks ago.


Roguelazer:
We do not bump versions from pull requests or from forks; I can do a release if you need one.

PR #117, opened 2020-12-22
parcels do not support listing and never have

Roguelazer:
It has never in the life of the company been possible to list parcels.

Fixes internal issue T28563.


Justintime50 (approved):
Haha yep, this needs to happen.

jontsai (approved):

PR #118, opened 2021-01-04
add github action

Roguelazer:
see internal task T28704 for details


.github/workflows/test.yaml
```diff
@@ -1,3 +1,10 @@
+name: 'CI'
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
Justintime50, line 6: Is this correct `yml` syntax here? Leaving the colon without anything inside (beneath) it?

Roguelazer: it's what the github docs say to do. I believe it's equivalent to saying `pull_request: ~` and should deserialize to

```
{"pull_request": None}
```

Justintime50: Perfect! Thanks

.github/workflows/test.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/checkout@v2
+      - name: set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.pythonversion }}
+      - name: install dependencies
+        run: |
+          python -m pip install six requests flake8
+          python -m pip install -r requirements-tests.txt
+          python -m pip install -e .
```
Justintime50, line 24: Won't `python -m pip install -e .` also be installing the requirements from `setup.py` and so installing `six` and `requests` separate above would be redundant?

Roguelazer: probably. I was just translating from the travis config, which copied that from the circle config, which got it from gods know where

Justintime50 (approved):
LGTM, would love to get flake8 into our test requirements so it's ready to use whenever including locally.

Justintime50:
Ah, can we also add the GitHub Actions README badge and swap it for the Travis badge?

Justintime50 (approved):
LGTM

Justintime50:
Looks like we'll need to remove Travis CI checks in the settings of the repo before this goes green.

PR #119, opened 2021-02-03
add new regenerate_rates method corresponding to the new rerate api

Roguelazer:
Starting soon in the future,`get_rates()` will always be read-only (as you would expect from a GET), and re-rating will be triggered through the new `regenerate_rates()` function. This adds the new function. No change has occurred to `get_rates()` yet.


jontsai (approved):
LGTM

PR #124, opened 2021-05-17
fix: stops appending smartrates to Shipment object

Justintime50:
Fixes Smartrate by not appending the results to a Shipment object anymore.

Previously we were appending the `results` list from Smartrates to the Shipment object by refreshing the record. We don't want to be doing this - instead we want to simply return the smartrate results without the accompanying shipment data.

[This was caught](https://github.com/EasyPost/easypost-ruby/pull/117#discussion_r632764921) in the Ruby library and others but I failed to see it here.

I was able to end-to-end test this to ensure it only returned the Smartrate `results` array and not the `Shipment` object in addition.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -733,8 +733,7 @@ class Shipment(AllResource, CreateResource):
         url = "%s/%s" % (self.instance_url(), "smartrate")
 
         response, api_key = requestor.request('get', url)
-        self.refresh_from(response, api_key)
-        return self
+        return response
 
     def buy(self, **params):
         requestor = Requestor(self._api_key)
--- a/tests/test_shipment.py
+++ b/tests/test_shipment.py
@@ -216,10 +216,10 @@ def test_smartrate(vcr):
 
     smartrates = shipment.get_smartrates()
     assert shipment.rates[0]['id'] == smartrates['result'][0]['id']
-    assert smartrates['result'][0]['time_in_transit']['percentile_50'] == 2
+    assert smartrates['result'][0]['time_in_transit']['percentile_50'] == 1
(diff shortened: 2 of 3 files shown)


easypost/__init__.py
```diff
@@ -733,8 +733,7 @@ def get_smartrates(self):
         url = "%s/%s" % (self.instance_url(), "smartrate")
 
         response, api_key = requestor.request('get', url)
-        self.refresh_from(response, api_key)
-        return self
```
pauloppenheim-easypost, line 737: Core of the change, return response not self, seems legit

tests/test_shipment.py
```diff
@@ -216,10 +216,10 @@ def test_smartrate(vcr):
 
     smartrates = shipment.get_smartrates()
     assert shipment.rates[0]['id'] == smartrates['result'][0]['id']
-    assert smartrates['result'][0]['time_in_transit']['percentile_50'] == 2
+    assert smartrates['result'][0]['time_in_transit']['percentile_50'] == 1
     assert smartrates['result'][0]['time_in_transit']['percentile_75'] == 2
-    assert smartrates['result'][0]['time_in_transit']['percentile_85'] == 3
+    assert smartrates['result'][0]['time_in_transit']['percentile_85'] == 2
```
pauloppenheim-easypost, line 221: Assuming this is just that the results changed based on the shipment defaults at time of call?

Justintime50: Yeah, re-recorded the cassette for good measure, these can change on each run.

easypost/__init__.py
Justintime50: Right

pauloppenheim-easypost (approved):

jontsai (approved):

PR #125, opened 2021-05-18
bump to v5.1.1 fix SmartRate response JSON format

jontsai:


Justintime50 (approved):

Roguelazer (changes requested):
this is missing `version.py`.

Justintime50 (approved):

setup.py
```diff
@@ -23,7 +24,7 @@
 
 setup(
     name='easypost',
-    version='5.1.0',
+    version='5.1.1',
```
pseudoben, line 27: It seems like it's too easy for this to get out of sync with version.py. Maybe this should be automatically populated?

pseudoben (approved):
Suggestion of auto-populating version from version.py is non-blocking.

setup.py
jontsai: Good suggestions. Perhaps a unit test or something.

An earlier version was trying to to some introspection, but it messed with packaging/setuptools.

Roguelazer (approved):

PR #129, opened 2021-06-10
chore: bump version for smartrate correction

Justintime50:
Bumps version to 5.1.2 for Smartrate fix (patch).

Not sure the best version bump here as we intended for Smartrate to not include the `result` key but it is a breaking change. Open to thoughts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v5.1.2 2021-06-10
+
+* Strips away the `result` key from SmartRate and simply returns an array of SmartRate objects
+
 ### 5.1.1 2021-05-18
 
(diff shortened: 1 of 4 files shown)


Sarahbear1926 (approved):

ejachang (approved):

jontsai:
For context here, we want to bump the patch version and release `5.1.2` (out of `major.minor.patch`) but this does break something over `5.1.1`,  but it’s technically a bug fix for something that should have never been released.

So we will want to look up which users have made `smartrate` requests using `5.1.1`, and just manually email them to give them a heads up to make updates to their code. This would have been in the last ~2 weeks.

PR #139, opened 2021-10-02
Encode put/post body as json instead of urlencoding

fiendish:
Second try (sorry). 

The urlencoding of post bodies breaks submission of shipment tax_identifiers, but that's ok because we should submit json bodies anyway.

I really do think it's just this, though the bulkiness of that custom encoder gives me pause. What's the story there? Anyway, tests pass, and I'm not experienced enough with the API yet to know whether there are any lurking gotchas.

Side note: I tested in a local google appengine environment and that worked too, but it's a bit beyond my scope to get the repo CI to run both.


Justintime50 (changes requested):
Looks like a good first pass, thanks for testing on the Google platform!

We'll want to remove the `_urlencode_list` function and ensure that the `Address.create` function works correctly still when we want to verify an address: https://www.easypost.com/docs/api#create-and-verify-addresses

There's other functions related to encoding such as `encode_list` that may not be necessary anymore which we'd want to investigate.

fiendish:
Good points. Coming up.

tests/cassettes/test_address_bytestring.yaml
```diff
@@ -1,17 +1,15 @@
 interactions:
 - request:
-    body: !!python/unicode address%5Bstate%5D=DELEGACI%C3%B3N+BENITO+JU%C3%A1REZ
+    body: !!python/unicode '{"address": {"state": "DELEGACI\u00f3N BENITO JU\u00e1REZ"}}'
```
Justintime50, line 3: Is this the correct behavior we are expecting here? We have a test `test_address_bytestring` to ensure these get created correctly, were we wanting it to converto to the actual letters or be as they show here?

easypost/__init__.py
```diff
@@ -217,8 +190,6 @@ def encode_none(cls, out, key, value):
     def _encode_inner(cls, params):
```
Justintime50, line 190: I don't think the `_encode_inner` is being called anymore, is that so? If not, we'll want to remove it; however, it was calling the encoder for `datetime`. Is that now getting properly encoded? Do we have a test for that? If so and `encode_datetime` isn't needed anymore, we could also remove that.

Similarly, I don't believe that `def encode` is being used anymore either and if so could be removed along with the `urlencode` function from `six`.

Justintime50 (changes requested):
Thanks for re-recording cassettes! That's super helpful.

A few questions and cleanup items still.

easypost/__init__.py
fiendish: > I don't think the _encode_inner is being called anymore, is that so?

It still does for GET requests via build_url -> encode.

> it was calling the encoder for datetime. Is that now getting properly encoded?

I only saw datetime arguments on GET requests, so those should still be ok since it's still being used there.

> Similarly, I don't believe that def encode is being used anymore either

See previous.

Justintime50: Ah perfect, must have just missed one of the calls. Thanks.

tests/cassettes/test_address_bytestring.yaml
fiendish: Bytestrings are getting passed through _utf8 as they were before in the urlencoder. 

```
>>> print(easypost._utf8(b'DELEGACI\xc3\xb3N BENITO JU\xc3\xa1REZ'))
DELEGACIóN BENITO JUáREZ

>>> print('DELEGACI\u00f3N BENITO JU\u00e1REZ')
DELEGACIóN BENITO JUáREZ
>>> 
```

Justintime50:
We seem to be covering most items here now, thanks for the work on this! I'll wait to approve until our group review.

ajnassar (dismissed):

ajnassar (approved):

fiendish:
`encode_url` feels overly bulky for what it's doing now, but I think I'd rather put refactoring (if any) that into a separate PR.

fiendish:
`encode_url` feels overly bulky for what it's doing now, but I think I'd rather put refactoring (if any) that into a separate PR.

easypost/__init__.py
```diff
@@ -214,11 +187,9 @@ def encode_none(cls, out, key, value):
         pass  # do not include None-valued params in request
 
     @classmethod
-    def _encode_inner(cls, params):
+    def encode_url(cls, params):
```
Justintime50, line 190: Yes, thank you for changing this name, this is much better.

easypost/__init__.py
```diff
@@ -240,7 +211,7 @@ def _encode_inner(cls, params):
                     pass
 
                 out.append((key, value))
-        return out
+        return urlencode(out)
```
Justintime50, line 214: Could we assign `urlencode(out)` to a variable and return a clean variable?

Justintime50 (approved):

easypost/__init__.py
fiendish: If you prefer. I tend to favor the opposite because I don't like extra indirection.

fiendish:
> I'd like @jontsai to review before we move on this one.

Speaking of, I also want to know whether any of #145 should get merged into this.

jontsai (approved):

PR #141, opened 2021-10-04
Make unit tests run in a local appengine environment

### LINKED ISSUE #140: Testing in local appengine environment causes non-test-failure errors
I encountered two issues testing locally with google appengine instead of requests lib:

1. Locally invoking urlfetch.fetch raises `*** AssertionError: No api proxy found for service "urlfetch"`
2. https://github.com/EasyPost/easypost-python/blob/8af4e9d7fa5d6f71d2cc4625ef546bab489d76e2/tests/test_shipment.py#L177 assumes that easypost.requests_session exists, which is only true when using requests lib.

fiendish:
Some small tweaks are needed to prevent non-test-error failures when testing with the appengine sdk instead of requests lib.

closes #140 

With the google cloud SDK installed and PYTHONPATH pointing at `google-cloud-sdk/platform/google_appengine/`, tests fail abruptly without these changes and pass with them.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,25 @@ TEST_API_KEY = os.environ['TEST_API_KEY']
 PROD_API_KEY = os.environ['PROD_API_KEY']
 
 
+def pytest_sessionstart(session):
+    # this is for local unit testing with google appengine, otherwise you get a
+    # 'No api proxy found for service "urlfetch"' response
+    try:
+        from google.appengine.ext import testbed
+        session.appengine_testbed = testbed.Testbed()
+        session.appengine_testbed.activate()
+        session.appengine_testbed.init_urlfetch_stub()
+    except ImportError:
+        # if the import fails then we're not using the appengine sdk, so just
+        # keep going without initializing the testbed stub
+        pass
+
+
+def pytest_sessionfinish(session, exitstatus):
+    if hasattr(session, "appengine_testbed"):
+        session.appengine_testbed.deactivate()
+
+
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -10,6 +10,23 @@
...
 
+def pytest_sessionstart(session):
+    # this is for local unit testing with google appengine, otherwise you get a
+    # 'No api proxy found for service "urlfetch"' response
+    try:
+        from google.appengine.ext import testbed
+        session.appengine_testbed = testbed.Testbed()
+        session.appengine_testbed.activate()
+        session.appengine_testbed.init_urlfetch_stub()
+    except Exception:
```
Justintime50, line 21: When would we arrive at this exception? Would we instead want to raise a message of sorts here instead of passing since we'd want the above to run correctly and not fail silently?

fiendish: The exception is for normal operation when _not_ running in a google appengine setup. This essentially mirrors the library selection mechanism in https://github.com/EasyPost/easypost-python/blob/8af4e9d7fa5d6f71d2cc4625ef546bab489d76e2/easypost/__init__.py#L24-L26

Want me to change to except ImportError instead?

Justintime50: Ah, if that's the case it may be advantageous to simply add a comment under the except and above the pass explaining how we'd get there. Your point of `not initialize the urlfetch testbed stub since we're not using the appengine sdk.` could be sufficient!

Justintime50 (approved):
Love it, thanks!

jontsai (approved):
LGTM

PR #142, opened 2021-10-04
Bump minimum required requests lib version slightly

### LINKED ISSUE #137: Adapter max_retries initialization won't work for older requests library that might be installed
https://github.com/EasyPost/easypost-python/blob/8af4e9d7fa5d6f71d2cc4625ef546bab489d76e2/setup.py#L10-L13

https://github.com/EasyPost/easypost-python/blob/8af4e9d7fa5d6f71d2cc4625ef546bab489d76e2/easypost/__init__.py#L34

setup.py specifies requests>=1.0.0 but the max_retries adapter initialization argument was not added until requests v2.26.0, so this line will error after error-free installation if 1.0.0<=requests<2.26.0 is installed.

Is 1.0.0 still an important version of requests lib to maintain support for?

fiendish:
The max_retries HTTPAdapter init argument was not added until requests
v1.2.0 so if we're going to use that initializer then we should bump the
minimum requests version from 1.0.0 to 1.2.3 (the last patch on 1.2).

closes #137


Justintime50 (approved):

setup.py
```diff
@@ -8,7 +8,7 @@
     from distutils.core import setup
 
 install_requires = [
-    'requests >= 1.0.0',
+    'requests >= 1.2.3',
```
jontsai, line 11: # Reviewer Note

Confirmed that `v1.2.3` is the highest patch version of `v1.2.x` here: https://pypi.org/project/requests/#history

jontsai (approved):
LGTM

Sarahbear1926 (approved):

PR #145, opened 2021-10-05
Slim down the query encoder

fiendish:
This removes extraneous encoder code (without losing important functionality of course). A few things have been renamed for clarity, and some class methods (encoding URL params) that I felt had no relation to the class have been moved to functions as well.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -160,6 +160,24 @@ def _utf8(value):
     return value
 
 
+def encode_url_params(params):
+    converted_params = []
+    for key, value in sorted(six.iteritems(params)):
+        if value is None:
+            continue  # don't add Nones to the query
+        elif isinstance(value, datetime.datetime):
+            value = int(time.mktime(value.timetuple()))  # to UTC timestamp
+        converted_params.append((key, value))
+    return urlencode(converted_params)
+
+
+def add_params_to_url(url, params):
+    encoded_params = encode_url_params(params)
+    if encoded_params:
+        return '%s?%s' % (url, encoded_params)
+    return url
+
+
 class Requestor(object):
     def __init__(self, local_api_key=None):
         self._api_key = local_api_key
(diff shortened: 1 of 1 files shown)


easypost/__init__.py
```diff
@@ -154,90 +153,45 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
 
 
 def _utf8(value):
-    if six.PY2:
-        # Python2's urlencode wants bytestrings, not unicode
```
Justintime50, line 158: # Reviewer Note

Talking with Avi, it appears that the Python2 encoding stuff may not be necessary here as tests pass without cassettes with all these changes.

easypost/__init__.py
```diff
@@ -154,90 +153,45 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
...
     @classmethod
     def _objects_to_ids(cls, param):
         if isinstance(param, Resource):
             return {'id': param.id}
         elif isinstance(param, dict):
-            out = {}
-            for k, v in six.iteritems(param):
-                out[k] = cls._objects_to_ids(v)
-            return out
+            return {k: cls._objects_to_ids(v) for k, v in six.iteritems(param)}
```
Justintime50, line 189: It could be advantageous to name these return values for easy code scanning?

easypost/__init__.py
```diff
@@ -531,19 +482,16 @@ def refresh(self):
...
-    def class_name(cls):
-        if cls == Resource:
-            raise NotImplementedError('Resource is an abstract class. '
-                                      'You should perform actions on its subclasses.')
-
-        cls_name = six.text_type(cls.__name__)
-        cls_name = cls_name[0:1] + re.sub(r'([A-Z])', r'_\1', cls_name[1:])
-        return "%s" % cls_name.lower()
+    def snakecase_name(cls):
+        """ Return the class name as snake_case.
```
Justintime50, line 486: Nit: Let's kill the extra space between `" Return`.

One-liner docstrings should also be on a single line.

https://www.python.org/dev/peps/pep-0257/

Justintime50 (commented):
Overall this looks good! Would love other Python fans to chime in.

easypost/__init__.py
fiendish: Or if it is necessary then we're not testing cases that matter!

easypost/__init__.py
fiendish: I'm very amenable if anyone has a name that isn't just a name for its own sake.
`{k: f(v) for k, v in dict.items()}` and `[f(v) for v in list]` are such common and simple idioms that a name would have to be very expressive to improve clarity here.

```
param_with_Resources_as_their_ids = {k: cls._objects_to_ids(v) for k, v in six.iteritems(param)}
return param_with_Resources_as_their_ids
```
? 😄

jontsai (dismissed):

Justintime50 (dismissed):

Justintime50 (approved):

jontsai (approved):

PR #147, opened 2021-11-02
Removes `_max_timeout` and instead uses a flat 60 second timeout for requests

Justintime50:
Removes `_max_timeout` and instead uses a flat 60 second timeout for requests to match other client libraries.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -14,7 +14,12 @@ __author__ = "EasyPost <oss@easypost.com>"
 __version__ = VERSION
 version_info = VERSION_INFO
 SUPPORT_EMAIL = "support@easypost.com"
+USER_AGENT = "EasyPost/v2 PythonClient/{0}".format(VERSION)
 TIMEOUT = 60
+
+# config
+api_key = None
+api_base = "https://api.easypost.com/v2"
 
 # use urlfetch as request_lib on google app engine, otherwise use requests
 request_lib = None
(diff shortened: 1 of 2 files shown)


easypost/__init__.py
```diff
@@ -61,8 +60,7 @@
 # config
 api_key = None
 api_base = "https://api.easypost.com/v2"
-# use our default timeout, or our max timeout if that is less
-timeout = min(60, _max_timeout)
+timeout = _platform_timeout
```
fiendish, line 63: I don't see any reason for a distinction between `_platform_timeout`/`timeout` to exist now. Can you remove this line and rename _platform_timeout to timeout on lines 26 and 28?

Justintime50: I initially did and put it back for fear that others would need/want the `urlfetch.get_default_fetch_deadline()` bit. In the off chance the GAE has a lower deadline, we'd want to get that instead.

fiendish: Leave it as `timeout = urlfetch.get_default_fetch_deadline() or TIMEOUT` and just don't redirect the variable name.

Justintime50: It is where it is because `urlfetch` isn't used by default (requests is). We only run that line of code if we import that module, otherwise (when not in a GAE context), we use the normal timeout.

Justintime50: Ah, thanks for clarifying offline. Updated the variable per your suggestion.

fiendish (approved):

PR #149, opened 2021-12-18
Use GitHub Actions to upload PyPi package when release created

nwithan8:
- Support for using GitHub Actions to trigger a release
- User will only ever need to edit `_info.py` (as needed)
- Copyright date (in code) will always remain current programmatically

Release Workflow Overview:
- User creates a release on GitHub (no file edits required beforehand)
- Making a release triggers the `publish_to_pypi` action:
  - Uses regex to replace VERSIONADDEDBYGITHUB in `easypost/_version.py` with the tag name used when creating release (use X.Y.Z semver only; PyPi does not support non-digits and will reject the upload, which will cause the Action to fail)
  - Uses `setup.py` to create the Python package from release, pulling info from `_info.py` and some environmental variables passed into the GitHub Actions runner (releases' tag name (i.e. X.Y.Z) and GitHub repo name (i.e. easypost/easypost-python))
  - Uses PYPI_USERNAME and PYPI_PASSWORD from GitHub Secrets to upload package to PyPi.org
- Package's `__author__` and `__version__` are set in `__init__.py` at runtime, using details from `_info.py`


nwithan8:
@Justintime50 not sure why the tests are failing. It seems like it's not installing the requirements properly, specifically `six`?

Justintime50:
There are some internal discussions we'll want to have about items like this before we move on them. Can we toss this into draft mode until that point?

Justintime50 (changes requested):
There are some internal discussions we'll want to have about items like this before we move on them. Can we toss this into draft mode until that point?

Justintime50:
This has been deprioritized for now, I'm going to close it. If ever we pick this back up, it'll be here waiting for us.

Issue #134, opened 2021-09-07
Error Handling Around Customs Items Not Exposing Proper Message


Justintime50:
# Summary

This will need some additional investigation; but it appears that when creating a custom item with a description that is longer than the max field length (~250 characters), the Python CL is returning an unhelpful `Wrong parameter type` error. In reality, it should be returning a message stating the max field length was exceeded or something along those lines. This info should be getting returned from the API but may not be getting exposed via this library.

## Acceptance Criteria

- [ ] Investigate why the `Wrong parameter type` error is occurring
- [ ] Investigate if this error occurs for other objects/fields or is only for this one field customs_item['description']
- [ ] Create a PR to fix the error handling to properly return the message

## Additional Info

<img width="1010" alt="Screen Shot 2021-09-07 at 1 28 06 PM" src="https://user-images.githubusercontent.com/39606064/132408298-a3e6cddc-e858-4548-ab31-dc4536aa9354.png">

A typical error response should look something like:

```python
{"error":{"code":"PARAMETER.INVALID_TYPE","message":"Wrong parameter type.","errors":[{"field":"weight","message":"must be a float"}]}}
```

jchen293:
The same error was thrown for the `quantity` field in the custom item object.

jchen293:
It's also happening in other objects as well. I believe here is the [code](https://github.com/EasyPost/easypost-python/blob/master/easypost/__init__.py#L328-L331) that is causing the issue. We are only displaying full error message when the exception is AttributeError. I have the PR in draft but ran into an issue with one of the unit tests.

Justintime50:
After looking into this, it appears that we were only exposing the `message` field when we raised an error when really there was an additional `code` and `errors` fields that were helpful to determine the true identity of an error. @jchen293 put up a PR to address this; however, it's a breaking change since it no longer unfurls the `message` key automatically and instead the user will be returned the full JSON error including the message field. This is a breaking change and will need a major version bump.

Justintime50:
After doing some internal testing, the default behavior of printing the `error` is to only print the `message` field; however, you could easily instead print the `error.http_body` and get the entire message. There are ways to work with the whole message then at that point. Closing as invalid.

PR #151, opened 2022-01-07
minor change on error handling

jchen293:
issue #134

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -65,7 +65,7 @@ except ImportError:
 
 class Error(Exception):
     def __init__(self, message=None, http_status=None, http_body=None, original_exception=None):
-        super(Error, self).__init__(message)
+        super(Error, self).__init__(http_body)
         self.message = message
         self.http_status = http_status
         self.http_body = http_body


tests/test_webhook.py
```diff
@@ -39,4 +39,4 @@ def test_webhooks(per_run_unique):
         easypost.Webhook.retrieve(webhook.id)
 
     assert exception_context.value.http_status == 404
-    assert exception_context.value.message == "The requested resource could not be found."
+    assert exception_context.value.message.get("message") == "The requested resource could not be found."
```
Justintime50, line 42: It's a code smell when we have to adjust tests to fit changed behavior. Tests are typically the "standard" the code should live by and when assertions change, it's an indicator that behavior is changing for our users too. For instance, users may be relying on the fact we unfurl the `message` key - with this change, we may inadvertently be introducing changed behavior around error handling which could break integrations.

I'm unsure this is the proper change here.

jchen293: I think you're right, another way to fix this is bypassing **http_body** instead of **message** in the Error class, so the Exception is going to print out the entire error message from API.

Justintime50:
After looking at this and testing, this is most likely the correct change; however, it's breaking change and will require a major version bump. We'll need to determine if there are other items we can include so we don't have a major version bump for just one thing.

Justintime50:
Closing as we won't be tweaking error handling soon and there are ways to get the whole error message via each integration.

PR #156, opened 2022-02-03
improve create_list for tracker class.

jchen293:
# what
Improve the **create_list** for our Python library Tracker class.
# why
Currently, the user must define a map that is named "trackers". An error can be thrown if the user mistypes the word. Instead, in our create_list method, we are going to create another map that has "trackers" as the key and user input as the value.
# how
Add another map in **create_list** method and define "trackers" as the key and store user input as the value.
# testing
Re-record the cassette and pass the unit test.

### DIFF
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -791,11 +791,11 @@ class PostageLabel(AllResource, CreateResource):
 
 class Tracker(AllResource, CreateResource):
     @classmethod
-    def create_list(cls, params={}, api_key=None):
+    def create_list(cls, trackers, api_key=None):
         requestor = Requestor(api_key)
         url = "%s/%s" % (cls.class_url(), "create_list")
-        new_params = {"trackers": params}
-        response, api_key = requestor.request("post", url, new_params)
+        new_params = {"trackers": trackers}
+        _, _ = requestor.request("post", url, new_params)
         return True
 
 
--- a/easypost/__init__.py
+++ b/easypost/__init__.py
@@ -794,7 +794,7 @@ class Tracker(AllResource, CreateResource):
     def create_list(cls, api_key=None, **params):
         requestor = Requestor(api_key)
         url = "%s/%s" % (cls.class_url(), "create_list")
-        newParams = {'trackers': params}
+        newParams = {"trackers": params}
         response, api_key = requestor.request("post", url, newParams)
         return True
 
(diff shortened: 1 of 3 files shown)


easypost/__init__.py
```diff
@@ -794,7 +794,8 @@ class Tracker(AllResource, CreateResource):
     def create_list(cls, api_key=None, **params):
         requestor = Requestor(api_key)
         url = "%s/%s" % (cls.class_url(), "create_list")
-        response, api_key = requestor.request("post", url, params)
+        newParams = {'trackers': params}
```
Justintime50, line 797: Python needs snake case variables, should be `new_params`.

tests/cassettes/test_tracker_create_list.yaml
```diff
@@ -1,8 +1,8 @@
 interactions:
 - request:
-    body: '{"trackers": {"0": {"tracking_code": "EZ1000000001"}, "1": {"tracking_code":
-      "EZ2000000002"}, "2": {"tracking_code": "EZ3000000003"}, "3": {"tracking_code":
-      "EZ4000000004"}}}'
+    body: !!python/unicode '{"trackers": {"params": {"1": {"tracking_code": "EZ2000000002"},
```
Justintime50, line 3: Why did this change to unicode?

tests/test_tracker.py
```diff
@@ -73,7 +73,7 @@ def test_tracker_interactions(per_run_unique):
 def test_tracker_create_list():
     """Tests that we can create a list of trackers in bulk."""
     trackers_list = easypost.Tracker.create_list(
-        trackers={
+        params={
```
Justintime50, line 76: The goal of this PR is to avoid the need to pass in a field at all, is that not possible with this implementation?

jchen293: We need a map to store those maps.

tests/cassettes/test_tracker_create_list.yaml
jchen293: Not sure yet. All I did is re-record the cassette.

easypost/__init__.py
```diff
@@ -791,10 +791,11 @@ class PostageLabel(AllResource, CreateResource):
 
 class Tracker(AllResource, CreateResource):
     @classmethod
-    def create_list(cls, api_key=None, **params):
+    def create_list(cls, params={}, api_key=None):
```
Justintime50, line 794: `api_key` is the first param to all functions in this library. Please adjust the ordering here.

Justintime50 (dismissed):

easypost/__init__.py
jchen293: If I swap back to the one we had, the user has to pass a **None** in the first parameter because we are no longer using keyword arguments in the parameters.

nwithan8 (approved):
LGTM

Justintime50 (approved):

PR #161, opened 2022-03-07
split __init__.py to multiple moduels.

jchen293:
work in progress


easypost/postage_label.py
```diff
@@ -0,0 +1,5 @@
+from easypost.resource import AllResource, CreateResource
+
+
+class PostageLabel(AllResource, CreateResource):
```
Justintime50, line 4: This should probably only inherit `Resource`

easypost/rate.py
```diff
@@ -0,0 +1,5 @@
+from easypost.resource import CreateResource
+
+
+class Rate(CreateResource):
```
Justintime50, line 4: This should probably only inherit `Resource`

easypost/resource.py
```diff
@@ -0,0 +1,284 @@
...
+class DeleteResource(Resource):
+    def delete(self, **params):
+        requestor = Requestor(self._api_key)
+        url = self.instance_url()
+        response, api_key = requestor.request("delete", url, params)
+        self.refresh_from(response, api_key)
+        return self
+
+
+class Requestor(object):
```
Justintime50, line 103: I'm still unsure how I feel about this class living in this module... will need to think on it.

easypost/resource.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def request(self, method, url, params=None, api_key_required=True):
+        if params is None:
+            params = {}
+        http_body, http_status, my_api_key = self.request_raw(method, url, params, api_key_required)
+        response = self.interpret_response(http_body, http_status)
+        return response, my_api_key
+
+    def request_raw(self, method, url, params=None, api_key_required=True):
+        from easypost import (
```
Justintime50, line 126: We should import everything at the top of a file when possible. Importing elsewhere (unless for a specific reason such as wrapping in a try/catch like we do for the `google` package is typically not the best option because it hides away the imports inside the logic.

easypost/resource.py
```diff
@@ -0,0 +1,284 @@
...
+            value = int(time.mktime(value.timetuple()))  # to UTC timestamp
+        converted_params.append((key, value))
+        return urlencode(converted_params)
+
+
+def add_params_to_url(url, params):
+    encoded_params = encode_url_params(params)
+    if encoded_params:
+        return "%s?%s" % (url, encoded_params)
+    return url
```
Justintime50, line 284: These 3 functions should probably live inside the `Requestor` class since that's the only class that uses them.

easypost/shipment.py
```diff
@@ -0,0 +1,84 @@
+from easypost.error import Error
+from easypost.resource import AllResource, CreateResource, Requestor
+
+
+class Shipment(AllResource, CreateResource):
+    @classmethod
+    def track_with_code(cls, api_key=None, **params):
```
Justintime50, line 7: This `track_with_code` function is super dead. Please remove it here and add a note to the CHANGELOG.

tests/conftest.py
```diff
@@ -1,5 +1,4 @@
 # setup for py.test
```
Justintime50, line 1: We can remove this useless comment since we're in here, `conftest` inherently sets up pytest.

Justintime50 (changes requested):
Great work on breaking everything out! No problems there

1. Fix the couple of small items called out
2. I need to think a bit about that `Requestor` class and if it should remain there or be moved somewhere else.

easypost/rate.py
jchen293: Nice catch, looks like we never fix this in the past

easypost/postage_label.py
jchen293: good call

easypost/resource.py
jchen293: I tried that, this would cause `partially initialized module 'easypost'` which most likely is circular import, in the `__init__.py` we have modules that are importing resource.py.

Justintime50: Ugh. For now, please add an comment above this import statement detailing why it's here. Something like `importing here to avoid circular imports` is fine.

easypost/resource.py
Justintime50: We were able to break this back out to its own module.

Justintime50 (approved):

nwithan8 (approved):

easypost/requestor.py
```diff
@@ -0,0 +1,193 @@
...
+from easypost.error import Error
+
+
+class Requestor(object):
+    def __init__(self, local_api_key=None):
+        self._api_key = local_api_key
+
+    @classmethod
+    def _objects_to_ids(cls, param):
+        if isinstance(param, EasyPostObject):
```
Justintime50, line 18: # Reviewer Note

This was the only code change throughout the whole PR: We swapped the isinstance of a `Resource` for `EasyPostObject` which should be safe since Resource inherits from EasyPostObject but doesn't extend it aside from adding functions. Making this change resolves a circular import.

fiendish:
All those "# Importing here to avoid circular imports" look entirely self-inflicted and could be avoided by not using the top level init as a dumping ground.

Justintime50:
> All those "# Importing here to avoid circular imports" look entirely self-inflicted and could be avoided by not using the top level init as a dumping ground.

Absolutely. Have suggestions on how @jchen293 can move forward to resolve that?

easypost/requestor.py
```diff
@@ -0,0 +1,193 @@
...
+
+        try:
+            result = requests_session.request(
+                method,
+                abs_url,
+                headers=headers,
+                data=data,
+                timeout=timeout,
+                verify=True,
+            )
```
fiendish, line 125: This is a kinda janky way of making requests for a few reasons (which you may not care about)...

1. The HTTP verb (`get`/`delete` vs `post`/`put`) has no true bearing on whether parameters should be assigned as URL query arguments or put in the message body. Gets can have json bodies, and posts can have URL parameters. This code is centrally mandating a relationship between them that is only true anecdotally, not logically.
2. Setting a `Content-Type` header to `application/json` when there's no message body doesn't hurt but it also isn't useful.
3. The requests library after version 2.4.2 has a mechanism already for setting the right header and doing the right thing for json messages by assigning the raw object to the `json` argument instead of assigning a string dump of the object to `data`, and the only reason this code doesn't use the more modern json mechanism is because of some decision from historic Python2 to accept raw bytes for addresses instead of strings and coerce them here with _utf8 instead of elsewhere. But all strings are unicode now in Python3 anyway and there's no reason for anyone to pass addresses in as raw bytes.
4. The requests library will also put your URL params in the right place for you without needing to manually call urlencode and concatenate strings by assigning to the `params` argument instead of to `data`.

easypost/requestor.py
```diff
@@ -0,0 +1,193 @@
...
+import platform
+import ssl
+import time
+from urllib.parse import urlencode
+
+from easypost.easypost_object import EasyPostObject
+from easypost.error import Error
+
+
+class Requestor(object):
```
fiendish, line 12: (and elsewhere) Python3 classes don't benefit from inheriting object.

fiendish:
> > All those "# Importing here to avoid circular imports" look entirely self-inflicted and could be avoided by not using the top level init as a dumping ground.
> 
> Absolutely. Have suggestions on how @jchen293 can move forward to resolve that?

I'd start by moving everything that isn't declared in __all__ out of the init file and then internally importing things from where they actually live in the tree instead of from the top level.

fiendish:
Honestly, I think that even declaring an __all__ at all is a bad sign of an interface gone bad.

Justintime50:
> Honestly, I think that even declaring an `__all__` at all is a bad sign of an interface gone bad, because nobody in their right mind ever imports `*`.

Haha you'd hope not! This brings up a good question - do we want an `__all__` in case someone does?

fiendish:
> > Honestly, I think that even declaring an `__all__` at all is a bad sign of an interface gone bad, because nobody in their right mind ever imports `*`.
> 
> Haha you'd hope not! This brings up a good question - do we want an `__all__` in case someone does?

Unless you're actually changing the list of exposed entities, having it doesn't come with any benefit.

Justintime50:
> > > Honestly, I think that even declaring an `__all__` at all is a bad sign of an interface gone bad, because nobody in their right mind ever imports `*`.
> > 
> > 
> > Haha you'd hope not! This brings up a good question - do we want an `__all__` in case someone does?
> 
> Unless you're actually reducing the set of exposed entities, having it doesn't come with any benefit.

Currently it doesn't expose a couple internal classes such as `EasyPostObject` which are getting imported here.

fiendish:
> Currently it doesn't expose a couple internal classes such as `EasyPostObject` which are getting imported here.

I would argue that if you're importing something into your top level init that is then not designated as exposed, you've done something wrong by importing it there in the first place.

tests/test_address.py
```diff
@@ -139,5 +139,6 @@ def test_address_bytestring():
     state = u"DELEGACI\xf3N BENITO JU\xe1REZ"
```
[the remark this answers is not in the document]
jchen293: This test will be removed in the next PR which is the test suite.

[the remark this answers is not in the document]
jchen293: This test will be removed in the next PR which is the test suite.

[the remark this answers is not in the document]
jchen293: This test was created before this PR and will be removed in the next PR which is the test suite.

tests/test_address.py
```diff
@@ -139,5 +139,6 @@ def test_address_bytestring():
...
     # fmt: on
 
-    address = easypost.Address.create(state=state.encode("utf-8"))
+    utfstring = state.encode("utf8")
```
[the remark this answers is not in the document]
jchen293: I add that variable for debugging initially and forgot to remove it, nonetheless, this test will be removed in the next PR which is the test suite.

easypost/requestor.py
jchen293: Thanks for the suggestion! For this PR we would like to keep it small and just shuffle around the code into separate modules. We can add an internal ticket to take a look into this and add another PR for it.

easypost/requestor.py
jchen293: Good call

jchen293:
> > > All those "# Importing here to avoid circular imports" look entirely self-inflicted and could be avoided by not using the top level init as a dumping ground.
> > 
> > 
> > Absolutely. Have suggestions on how @jchen293 can move forward to resolve that?
> 
> I'd start by moving everything that isn't declared in `__all__` out of the init file and then internally importing things from where they actually live in the tree instead of from the top level. The only good reason to put things into init is to expose a flat interface for end users that doesn't require referencing chains of submodules, but the library itself shouldn't constrain itself to using the public interface for private work.

This has been fixed, I have removed the `EasyPostObject` from the __init__.py file.

.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 120
 extend-ignore = E203
+per-file-ignores = __init__.py:F401
```
fiendish, line 4: missing newline

fiendish:
> This has been fixed, I have removed the `EasyPostObject` from the **init**.py file.

I still see a bunch of local imports `from easypost` in the requestor class though?

jchen293:
> > This has been fixed, I have removed the `EasyPostObject` from the **init**.py file.
> 
> I still see a bunch of local imports `from easypost` in the requestor class though?

`Requestor` class need `Error` and `EasyPostObject` class

fiendish (approved):
Stellar job. Please of course exercise all the objects. :D

jontsai (approved):
LFTM

nwithan8 (approved):

PR #162, opened 2022-03-14
Test suite

jchen293:
This PR adds full test coverage to the library by testing everything "reasonable". The test coverage is 78%

Screen Shot 2022-03-15 at 1 23 37 PM


tests/address_test.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.vcr()
+def test_address_retrieve():
+    address = easypost.Address.create(**fixture.basic_address())
+
+    retrieved_address = easypost.Address.retrieve(address.id)
+
+    assert isinstance(retrieved_address, easypost.Address)
+    assert retrieved_address.id == address.id
```
jchen293, line 35: # reviewer note:

We can't compare instances here because we have to implement the method [__eq__](https://docs.python.org/3/reference/datamodel.html#object.__eq__) in each class

nwithan8: We could probably do some kind of "loop over every property of Object A, get the corresponding property on Object B, compare", but I think that would introduce some funky stuff. This works for now.

tests/batch_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+@pytest.mark.vcr()
+def test_batch_create_scanform():
+    shipment_data = fixture.one_call_buy_shipment()
+
+    batch = easypost.Batch.create(shipments=[shipment_data])
+    batch.buy()
+
+    # Uncomment the following line if you need to re-record the cassette
+    time.sleep(5)  # Wait enough time for the batch to process buying the shipment
```
nwithan8, line 69: If we've recorded the cassette already, we'll want to comment this out.

tests/batch_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+@pytest.mark.vcr()
+def test_batch_label():
+    shipment_data = fixture.one_call_buy_shipment()
+
+    batch = easypost.Batch.create(shipments=[shipment_data])
+    batch.buy()
+
+    # Uncomment the following line if you need to re-record the cassette
+    time.sleep(5)  # Wait enough time for the batch to process buying the shipment
```
nwithan8, line 100: Comment this out if we've recorded already

tests/carrier_account_test.py
```diff
@@ -0,0 +1,75 @@
...
+            "account_number": "A1A1A1",
+            "user_id": "USERID",
+            "password": "PASSWORD",
+            "access_license_number": "ALN",
+        },
+    }
+
+
+@pytest.mark.vcr()
+def test_carrier_account_create(prod_api_key):
```
nwithan8, line 20: Is prod key actually being used in here? Check for all these tests

nwithan8 (changes requested):

tests/batch_test.py
jchen293: oops, nice catch!

tests/carrier_account_test.py
jchen293: Yes, the default API key is using the test key, we need to pass in prod_api_key for carrier account testing since it require prod key.

nwithan8: What I'm saying is, it doesn't seem like the prod key is actually getting passed along to the inner functions to actually be used? Or am I missing some syntactical sugar?

jchen293: `prod_api_key` is coming from `conftest.py` as a fixture. Check out this doc: https://docs.pytest.org/en/latest/how-to/fixtures.html

nwithan8 (dismissed):

tests/address_test.py
Justintime50: I played briefly with this. Only comparing IDs isn't ideal when we can reliably compare the entire object (various bugs exposed themselves due to comparing whole objects in other libs). I also implemented `__eq__` in a class to try and didn't have luck, though I may have done so incorrectly on my first attempt.

My suggestion is we try to compare whole objects. I'd suggest doing a bit of extra digging here to see what our options are and give it a try on a couple objects.

easypost/requestor.py
```diff
@@ -117,9 +117,10 @@ def requests_request(self, method, abs_url, headers, params):
             http_body = result.text
             http_status = result.status_code
         except Exception as e:
+            breakpoint()
```
Justintime50, line 120: Please double check your own code prior to pushing it (self-review each file before you commit in your IDE), stuff like this should never make it into the repo.

easypost/requestor.py
```diff
@@ -117,9 +117,10 @@ def requests_request(self, method, abs_url, headers, params):
...
             raise Error(
                 "Unexpected error communicating with EasyPost. If this "
-                f"problem persists please let us know at {SUPPORT_EMAIL}.",
+                f"problem persists please let us know at {SUPPORT_EMAIL}...",
```
Justintime50, line 123: Let's leave this as it was.

tests/insurance_test.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.mark.vcr()
+def test_insurance_create():
+    shipment = easypost.Shipment.create(**fixture.one_call_buy_shipment())
+
+    insurance = easypost.Insurance.create(
+        to_address=fixture.basic_address(),
+        from_address=fixture.basic_address(),
+        tracking_code=shipment.tracking_code,
+        carrier=fixture.usps(),
+        amount="100",
```
Justintime50, line 16: Nit: You could use a fixture for this instead (see the Node PR for an example)

tests/insurance_test.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.mark.vcr()
+def test_insurance_retrieve():
+    shipment = easypost.Shipment.create(**fixture.one_call_buy_shipment())
+
+    insurance = easypost.Insurance.create(
+        to_address=fixture.basic_address(),
+        from_address=fixture.basic_address(),
+        tracking_code=shipment.tracking_code,
+        carrier=fixture.usps(),
+        amount="100",
```
Justintime50, line 33: Nit: You could use a fixture for this instead (see the Node PR for an example)

tests/refund_test.py
```diff
@@ -0,0 +1,42 @@
...
+import tests.fixture as fixture
+
+
+@pytest.mark.vcr()
+def test_refund_create():
+    shipment = easypost.Shipment.create(**fixture.one_call_buy_shipment())
+
+    retrieved_shipment = easypost.Shipment.retrieve(
+        shipment
+    )  # We need to retrieve the shipment so that the tracking_code has time to populate
```
Justintime50, line 13: Comments like this may read better if they came just before the code (that way they aren't formatted weird due to `Black`)

tests/report_test.py
```diff
@@ -0,0 +1,134 @@
...
+    except Exception:
+        assert Exception
+
+
+@pytest.mark.vcr()
+def test_report_all_no_type():
+    try:
+        _ = easypost.Report.all()
+    except Exception:
+        assert Exception
```
Justintime50, line 134: These tests aren't necessary. They were included in PHP for example because PHP had specific branching logic to that should have been tested whereas the Python lib does not have that.

Delete the cassettes along with this so they don't stay in the repo.

tests/shipment_test.py
```diff
@@ -0,0 +1,154 @@
...
+@pytest.mark.vcr()
+def test_shipment_convert_label():
+    shipment = easypost.Shipment.create(**fixture.full_shipment())
+    shipment.buy(rate=shipment.lowest_rate())
+    shipment.label(file_format="ZPL")
+
+    assert shipment.postage_label.label_zpl_url is not None
+
+
+# If the shipment was purchased with a USPS rate, it must have had its insurance set to `0` when bought
```
Justintime50, line 69: In Python, these large block comments should instead be docblocks, please move them just below the test function name and wrap them in trouble quotes. This is true for all the test comments that are at the "function" level.

Justintime50 (changes requested):
- The convention in Python is to name test files as `test_xyz`, you have them the opposite order. Please rename all the test files so that they start with the word `test` followed by an underscore and then the name of the object being tested.

- Please also replace the two occurrences of `@pytest.yield_fixture` in the `conftest.py` file with `@pytest.fixture` as this is now deprecated and throwing warnings on test runs.

- You can also remove the `per_run_unique` fixture from `conftest.py`, I don't think we need/use it anymore

tests/easypost_test.py
```diff
@@ -0,0 +1,9 @@
+import easypost.requestor as requestor
+
+
+def test_timeout():
+    assert requestor.timeout == 60
+
+    requestor.timeout = 1
```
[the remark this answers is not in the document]
jchen293: I just did, this is resolved.

tests/address_test.py
jchen293: I have created a helper function that is going to compare each attribute recursively. We do have to convert the instance to dictionary first for the comparsion.

tests/user_test.py
```diff
@@ -0,0 +1,82 @@
...
+def test_user_retrieve(prod_api_key):
+    user = easypost.User(fixture.child_user_id())
+
+    assert isinstance(user, easypost.User)
+    assert str.startswith(user.id, "user_")
+
+
+@pytest.mark.vcr()
+def test_user_retrieve_me(prod_api_key):
+    user = easypost.User(fixture.child_user_id())
```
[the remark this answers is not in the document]
jchen293: Oops, I forgot to change it after copy-paste.

tests/carrier_account_test.py
```diff
@@ -0,0 +1,75 @@
...
+    assert carrier_account.description == test_description
+
+
+@pytest.mark.vcr()
+def test_carrier_account_delete(prod_api_key):
+    carrier_account = easypost.CarrierAccount.create(**generate_carrier_account())
+
+    response = carrier_account.delete()
+
+    assert response is not None
```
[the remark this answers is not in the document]
jchen293: I think Python lib is a little different, the `delete()` method is going to refresh the object and does not return anything from the endpoint. I will just remove this assertion.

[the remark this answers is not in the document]
Justintime50: Erm... if the object gets deleted then it can't return an object haha...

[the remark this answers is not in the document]
jchen293: I just realized there is some issue with `refresh_from()` it's being called from the `delete()` method. The object is actually never updated to the empty response. Looking into this now.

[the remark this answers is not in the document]
jchen293: Adding an if statement to check if the response is empty in the `delete()` method.

tests/insurance_test.py
jchen293: A helper function is created on the top of test_insurance.py

tests/report_test.py
jchen293: removed

tests/address_test.py
jchen293: My own customized helper function didn't work well because the response object can have different types of attributes. Spent a few hours trying to make my helper functions work but later realized we can use [deepdiff](https://pypi.org/project/deepdiff/) to compare two instances if that's okay?

Justintime50: I'd rather not introduce another dependency if we can get away without it. Thinking on this a bit, I believe I may have seen an implementation of something like this from @jontsai in the past (that we use internally) that may not even be that many lines of code that we could snag and add here. We can ask him about this.

tests/test_pickup.py
```diff
@@ -1,148 +1,56 @@
...
-import time
-
 import pytest
-import pytz
 
 import easypost
 
-ONE_DAY = datetime.timedelta(days=1)
 
+def generate_pickup(one_call_buy_shipment, basic_pickup):
```
Justintime50, line 6: Let's call this `create_pickup` instead, it more clearly defines what's happening.

tests/test_pickup.py
```diff
@@ -1,148 +1,56 @@
...
-        instructions="Special pickup instructions",
-    )
-
-    assert pickup.pickup_rates != [], pickup.messages
-
-    pickup.buy(carrier=pickup.pickup_rates[0].carrier, service=pickup.pickup_rates[0].service)
+def test_pickup_buy(usps, one_call_buy_shipment, basic_pickup):
+    pickup = generate_pickup(one_call_buy_shipment, basic_pickup)
+
+    bought_pickup = pickup.buy(carrier=usps, service="Nextday")
```
Justintime50, line 38: This should be `NextDay` (capital D) and could be a fixture just like the `usps_service` is (call it `pickup_service`)

tests/test_shipment.py
```diff
@@ -1,228 +1,157 @@
...
     assert smartrates[0]["time_in_transit"]["percentile_85"] is not None
     assert smartrates[0]["time_in_transit"]["percentile_90"] is not None
     assert smartrates[0]["time_in_transit"]["percentile_95"] is not None
     assert smartrates[0]["time_in_transit"]["percentile_97"] is not None
     assert smartrates[0]["time_in_transit"]["percentile_99"] is not None
+
+
+@pytest.mark.vcr()
+def test_shipment_create_empty_objects(basic_shipment):
+    shipment_data = basic_shipment
```
Justintime50, line 116: You also need to create this shipment with an empty list for `customs_info` to test we don't break things with empty lists. You'll then need to assert that.

tests/test_shipment.py
```diff
@@ -1,228 +1,157 @@
...
+
+
+@pytest.mark.vcr()
+def test_shipment_create_with_ids(basic_address, basic_parcel):
+    from_address = easypost.Address.create(**basic_address)
+    to_address = easypost.Address.create(**basic_address)
+    parcel = easypost.Parcel.create(**basic_parcel)
+
+    shipment = easypost.Shipment.create(
+        from_address={"id": from_address.id}, to_address={"id": to_address.id}, parcel={"id": parcel.id}
```
Justintime50, line 149: Please always use trailing commas, Black will format this accordingly when you do so (breaking these up to their own lines for better readability).

```suggestion
        from_address={"id": from_address.id}, to_address={"id": to_address.id}, parcel={"id": parcel.id},
```

tests/test_tracker.py
```diff
@@ -1,85 +1,46 @@
...
-    assert trackers2["has_more"] is not None
-    assert trackers2["trackers"][0].id == tracker3.id  # Should be the same as the id for tracker3
+    assert len(trackers_array) <= page_size
+    assert trackers["has_more"] is not None
+    assert all(isinstance(tracker, easypost.Tracker) for tracker in trackers_array)
 
 
 @pytest.mark.vcr()
 def test_tracker_create_list():
-    """Tests that we can create a list of trackers in bulk."""
```
Justintime50, line 74: This comment was valid and should be left.

tests/test_tracker.py
```diff
@@ -1,85 +1,46 @@
...
-    success = easypost.Tracker.create_list(
+    response = easypost.Tracker.create_list(
         {
             "0": {"tracking_code": "EZ1000000001"},
-            "1": {"tracking_code": "EZ2000000002"},
-            "2": {"tracking_code": "EZ3000000003"},
-            "3": {"tracking_code": "EZ4000000004"},
```
Justintime50, line 80: This line was valid and could be left.

tests/test_tracker.py
```diff
@@ -1,85 +1,46 @@
...
+            "1": {"tracking_code": "EZ1000000002"},
+            "2": {"tracking_code": "EZ1000000003"},
         }
     )
 
-    # This endpoint/method does not return anything, just make sure the request doesn't fail
```
Justintime50, line 84: This comment was valid and should be left.

easypost/easypost_object.py
```diff
@@ -183,6 +183,9 @@ def __repr__(self):
     def __str__(self):
         return self.to_json(indent=2)
 
+    def __eq__(self, other):
```
nwithan8, line 186: I'd recommend putting a check here, like 
```python
if not isinstance(other, EasyPostObject):
     return False
````
to short-circuit the function if we're not comparing a non-EasyPostObject to this clearly EasyPostObject.

File: tests/test_user.py [hunk 23]
```diff
@@ -1,45 +1,81 @@
...
+    assert str.startswith(user.id, "user_")
+    assert user.phone == test_phone
 
 
 @pytest.mark.vcr()
-def test_child_user_create(prod_api_key):
-    # Create an address and then verify some fields to test whether it was created just fine.
-    child_user = easypost.User.create(name="Python All-Things-Testing")
+@pytest.mark.skip(
+    reason="Due to our inability to create child users securely,"
```
Justintime50, line 48: You'll need a space after the comma or before the next `we`, when concatenating strings it'll slap them together without a space unless you add one yourself.

tests/test_webhook.py
```diff
@@ -1,42 +1,52 @@
...
-    expected_url = "http://example.com/{0}".format(per_run_unique)
-
-    # Create a webhook
-    webhook = easypost.Webhook.create(url=url)
-    assert webhook.id is not None
-    assert webhook.mode == "test"
-    assert webhook.url == expected_url
-    assert webhook.disabled_at is None
+def test_webhook_create():
+    webhook = easypost.Webhook.create(url="http://example.com")
```
Justintime50, line 8: Nit: You could make the webhook url a fixture like the others so you can reuse it and if ever we needed to switch it it's easier.

Justintime50 (changes requested):
Once all the below has been addressed, I'll be good to stamp. Looked through the whole thing.

Please also remove the `pytz` dev dependency and associated TODO since it's no longer needed with this overhaul.

tests/test_order.py
```diff
@@ -1,95 +1,52 @@
...
+
+
+@pytest.mark.vcr()
+def test_order_retrieve(basic_address, basic_shipment):
+    order = create_order(basic_address, basic_shipment)
+
+    retrieved_order = easypost.Order.retrieve(order.id)
+
+    assert isinstance(retrieved_order, easypost.Order)
+    assert retrieved_order.id == order.id
```
nwithan8, line 30: Add a note to the `retrieve` test in `test_order`, `test_insurance`, `test_batch` and `test_tracker`:
```
# status changes between creation and retrieval, so we can't compare the whole object
```

nwithan8 (changes requested):

tests/test_order.py
Justintime50: 100% agree and thought to say the same, but didn't as we didn't do this in the other libs (but should have).

tests/test_address.py
```diff
@@ -1,144 +1,81 @@
...
-
-    assert len(exception["error"]["errors"]) >= 2
-    assert exception["error"]["errors"][0]["message"] == "Address not found"
-    assert exception["error"]["errors"][1]["message"] == "House number is missing"
+def test_address_create_verify(incorrect_address_to_verify):
+    """
+    Test creating a verified address.
+    We purposefully pass in slightly incorrect data to get the corrected address back once verified.
+    """
+    address = easypost.Address.create(**incorrect_address_to_verify)
```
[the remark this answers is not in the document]
jchen293: I will remove this function since the testing duplicate to the one you pointed out in the above comment: https://github.com/EasyPost/easypost-python/pull/162#discussion_r829439481

tests/test_carrier_account.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+
+import easypost
+
+
+def generate_carrier_account():
```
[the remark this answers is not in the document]
jchen293: We had this discussion yesterday. Since this helper function is only being used in `test_carrier_account.py`, we should be putting it in `confest.py`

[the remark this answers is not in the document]
Justintime50: I misunderstood what you were asking. Because this is `static data`, it belongs in `conftest.py` along with the other fixtures of static data. If this function actually did something (aside from returning a dictionary) and was only used here; then yes, having this helper function in this file would make sense but that's not the case here.

tests/test_shipment.py
jchen293: The assertions are correct and passed, I didn't see why we need it since it's already null. Anyway, I will add this back.

tests/test_tracker.py
jchen293: I was using the test input example from other libs to make it consistent, the ` "3": {"tracking_code": "EZ4000000004"},` is unnecessary, let's keep what I have now.

tests/test_user.py
jchen293: thanks for letting me know that!

tests/test_carrier_account.py
```diff
@@ -0,0 +1,75 @@
...
+        },
+    }
+
+
+@pytest.mark.vcr()
+def test_carrier_account_create(prod_api_key):
+    carrier_account = easypost.CarrierAccount.create(**generate_carrier_account())
+
+    assert isinstance(carrier_account, easypost.CarrierAccount)
+    assert str.startswith(carrier_account.id, "ca_")
```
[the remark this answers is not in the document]
Justintime50: Nah, there is a `type` property on this object and it should be `UpsAccount`.

tests/test_shipment.py
Justintime50: I understand these assertions are correct. I'm asking that **in addition** to these assertions you already have, you should also pass an empty list to `customs_info`. This tests purpose is to check that we encode data to the API correctly. We had big problems with the PHP encoding because we didn't test enough edge cases (users frequently pass us empty lists) and so we should test/guard against that by having a test that does this as we do in all our other libs.

jchen293: Thanks for the explanation, I have added an empty list to the customs_info field.

tests/test_carrier_account.py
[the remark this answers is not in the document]
jchen293: Make sense, I was following the PHP test suite. I will add this assertion.

tests/test_batch.py
```diff
@@ -1,92 +1,106 @@
...
+
+@pytest.mark.vcr()
+def test_batch_retrieve(one_call_buy_shipment):
+    batch = easypost.Batch.create(shipments=[one_call_buy_shipment])
+
+    retrieved_batch = easypost.Batch.retrieve(batch.id)
+
+    assert isinstance(batch, easypost.Batch)
+    assert retrieved_batch.id == batch.id
+    # status changes between creation and retrieval, so we can't compare the whole object
```
Justintime50, line 23: A comment after the code its referencing is confusing. Please always have comments either come right before the line they are referencing or on the same line (depending on length and context.)

tests/test_carrier_account.py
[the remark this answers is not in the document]
Justintime50: Yeah, as it was the first one done, there is a thing here and there that were only discovered later.

tests/test_carrier_account.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.vcr()
+def test_carrier_account_retrieve(prod_api_key, basic_carrier_account):
+    carrier_account = easypost.CarrierAccount.create(**basic_carrier_account)
+
+    retrieved_carrier_account = easypost.CarrierAccount.retrieve(carrier_account.id)
+
+    assert isinstance(retrieved_carrier_account, easypost.CarrierAccount)
+    assert carrier_account.id == retrieved_carrier_account.id
+    # status changes between creation and retrieval, so we can't compare the whole object
```
Justintime50, line 25: Comment before the line it's referencing.

tests/test_easypost.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_timeout():
+    assert requestor.timeout == 60
+
+    requestor.timeout = 1
+
+    assert requestor.timeout == 1
+
+    requestor.timeout = 60
```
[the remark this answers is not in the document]
Justintime50: No comment was added here but it was marked resolved. Please only mark items resolved if they actually get resolved.

tests/test_insurance.py
```diff
@@ -1,64 +1,44 @@
...
+    insurance_data = basic_insurance
+    insurance_data["tracking_code"] = shipment.tracking_code
+
+    insurance = easypost.Insurance.create(**insurance_data)
+
+    retrieved_insurance = easypost.Insurance.retrieve(insurance.id)
+
+    assert isinstance(retrieved_insurance, easypost.Insurance)
+    assert insurance.id == retrieved_insurance.id
+    # status changes between creation and retrieval, so we can't compare the whole object
```
Justintime50, line 33: Comment before code it's referencing.

tests/test_order.py
```diff
@@ -1,95 +1,53 @@
...
+    # status changes between creation and retrieval, so we can't compare the whole object
```
Justintime50, line 31: Comment before code it's referencing.

tests/test_shipment.py
```diff
@@ -1,228 +1,160 @@
...
+    shipment_data["options"] = None
+    shipment_data["tax_identifiers"] = None
+    shipment_data["reference"] = ""
+
+    shipment = easypost.Shipment.create(**shipment_data)
+
+    assert isinstance(shipment, easypost.Shipment)
+    assert str.startswith(shipment.id, "shp_")
+    assert shipment.options is not None
+    assert shipment.customs_info is None
```
Justintime50, line 127: I apologize if I mispoke here. The test should be an empty `customs_items` nested list (see PHP here: https://github.com/EasyPost/easypost-php/blob/master/test/EasyPost/ShipmentTest.php#L237).

We should then assert against the top level `tax_identifiers` (missing) and assert the nested `customs_items`. This way, if we ever changed encoding in this lib, it'd catch the regression because we are testing all sorts of things (empty strings, empty objects - top level and nested, objects that are `None`, etc)

tests/test_tracker.py
```diff
@@ -1,85 +1,49 @@
...
+    tracker = easypost.Tracker.create(tracking_code="EZ1000000001")
 
-    assert tracker2.id == tracker.id  # Should be the same as above
+    retrieved_tracker = easypost.Tracker.retrieve(tracker.id)
 
-    # retrieve all trackers by tracking_code
-    trackers = easypost.Tracker.all(tracking_code=tracking_code)
+    assert isinstance(retrieved_tracker, easypost.Tracker)
+    assert retrieved_tracker.id == tracker.id
+    # status changes between creation and retrieval, so we can't compare the whole object
```
Justintime50, line 23: Comment before code referencing it.

Justintime50 (changes requested):
Really close, a couple small last things.

tests/test_shipment.py
jchen293: My assertions are correct as well as the order, do you mean assert `tax_identifiers` is null which I don't see other libs have this yet.

jchen293: We will need to patch node.js lib as well for this one: https://github.com/EasyPost/easypost-node/blob/master/test/resources/shipment.test.js#L34

nwithan8 (approved):

nwithan8:
LGTM, thanks for working so hard on this!

tests/test_shipment.py
Justintime50: Yes please, add the assertion requested here.

jchen293: I will create a ticket in the backlog to add the assertion in all 7 libs so I don't dismiss Nate's approval again.

Justintime50 (approved):
Nice work!

PR #163, opened 2022-03-15
Add parameter and return type hints to all functions

nwithan8:
By providing type hints, we can make it our code semi-self-documenting, and a lot friendlier to those using IDEs (or really anyone). Users will now know what type each parameter of a function is supposed to be, as well as the final type that the function will return.

This PR includes:
- Parameter and return type hints for all functions (passed `MyPy` analysis)
- Explicit parameter assignment (i.e. `my_function(parameter=this, number=that)`). This makes it clear what parameter each  value is being mapped to when a function is called, and greatly improves the readability of our code base.
- A few minor tweaks to functions to improve bad logic, make naming/structure consistent, and fix loosely-typed functions

### DIFF
--- a/easypost/address.py
+++ b/easypost/address.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict, Optional
+
 from easypost.easypost_object import convert_to_easypost_object
 from easypost.requestor import Requestor
 from easypost.resource import AllResource, CreateResource
@@ -5,9 +7,15 @@ from easypost.resource import AllResource, CreateResource
 
 class Address(AllResource, CreateResource):
     @classmethod
-    def create(cls, api_key=None, verify=None, verify_strict=None, **params):
+    def create(
+        cls,
+        api_key: Optional[str] = None,
+        verify: Optional[Dict[str, Any]] = None,
+        verify_strict: Dict[str, Any] = None,
+        **params
+    ) -> "Address":
         """Create an address."""
-        requestor = Requestor(api_key)
+        requestor = Requestor(local_api_key=api_key)
         url = cls.class_url()
 
         wrapped_params = {cls.snakecase_name(): params}
@@ -18,45 +26,45 @@ class Address(AllResource, CreateResource):
                 value = [value]
             wrapped_params[key] = value
 
-        response, api_key = requestor.request("post", url, wrapped_params)
-        return convert_to_easypost_object(response, api_key)
+        response, api_key = requestor.request(method="post", url=url, params=wrapped_params)
+        return convert_to_easypost_object(response=response, api_key=api_key)
 
     @classmethod
-    def create_and_verify(cls, api_key=None, carrier=None, **params):
+    def create_and_verify(cls, api_key: Optional[str] = None, carrier: Optional[str] = None, **params) -> "Address":
         """Create and verify an address."""
-        requestor = Requestor(api_key)
+        requestor = Requestor(local_api_key=api_key)
         url = "%s/%s" % (cls.class_url(), "create_and_verify")
 
         wrapped_params = {cls.snakecase_name(): params, "carrier": carrier}
-        response, api_key = requestor.request("post", url, wrapped_params)
+        response, api_key = requestor.request(method="post", url=url, params=wrapped_params)
 
         response_address = response.get("address", None)
         response_message = response.get("message", None)
 
         if response_address is not None:
-            verified_address = convert_to_easypost_object(response_address, api_key)
+            verified_address = convert_to_easypost_object(response=response_address, api_key=api_key)
             if response_message is not None:
                 verified_address.message = response_message
                 verified_address._immutable_values.update("message")
             return verified_address
         else:
-            return convert_to_easypost_object(response, api_key)
+            return convert_to_easypost_object(response=response, api_key=api_key)
 
-    def verify(self, carrier=None):
+    def verify(self, carrier: Optional[str] = None) -> "Address":
         """Verify an address."""
-        requestor = Requestor(self._api_key)
+        requestor = Requestor(local_api_key=self._api_key)
         url = "%s/%s" % (self.instance_url(), "verify")
         params = {"carrier": carrier}
-        response, api_key = requestor.request("get", url, params=params)
+        response, api_key = requestor.request(method="get", url=url, params=params)
 
         response_address = response.get("address", None)
         response_message = response.get("message", None)
 
         if response_address is not None:
-            verified_address = convert_to_easypost_object(response_address, api_key)
+            verified_address = convert_to_easypost_object(response=response_address, api_key=api_key)
             if response_message is not None:
                 verified_address.message = response_message
                 verified_address._immutable_values.update("message")
             return verified_address
         else:
-            return convert_to_easypost_object(response, api_key)
+            return convert_to_easypost_object(response=response, api_key=api_key)
--- a/easypost/easypost_object.py
+++ b/easypost/easypost_object.py
@@ -1,9 +1,12 @@
 import json
+from typing import Any, Dict, List, Optional
 
 import easypost
 
 
-def convert_to_easypost_object(response, api_key, parent=None, name=None):
+def convert_to_easypost_object(
+    response: Dict[str, Any], api_key: Optional[str] = None, parent: object = None, name: Optional[str] = None
+):
     """Convert a response to an EasyPost object."""
     types = {
         "Address": easypost.Address,
@@ -63,7 +66,7 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
     }
 
     if isinstance(response, list):
-        return [convert_to_easypost_object(r, api_key, parent) for r in response]
+        return [convert_to_easypost_object(response=item, api_key=api_key, parent=parent) for item in response]
     elif isinstance(response, dict):
         response = response.copy()
         cls_name = response.get("object", EasyPostObject)
@@ -74,13 +77,20 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
             cls = prefixes.get(cls_id[0 : cls_id.find("_")], EasyPostObject)
         else:
             cls = EasyPostObject
-        return cls.construct_from(response, api_key, parent, name)
+        return cls.construct_from(values=response, api_key=api_key, parent=parent, name=name)
     else:
         return response
 
 
 class EasyPostObject(object):
-    def __init__(self, easypost_id=None, api_key=None, parent=None, name=None, **params):
+    def __init__(
+        self,
+        easypost_id: Optional[str] = None,
+        api_key: Optional[str] = None,
+        parent: object = None,
+        name: Optional[str] = None,
+        **params,
+    ):
         self.__dict__["_values"] = set()
         self.__dict__["_unsaved_values"] = set()
         self.__dict__["_transient_values"] = set()
@@ -95,7 +105,7 @@ class EasyPostObject(object):
         if easypost_id:
             self.id = easypost_id
 
-    def __setattr__(self, k, v):
+    def __setattr__(self, k, v: Any) -> None:
         self.__dict__[k] = v
 
         if k not in self._immutable_values:
@@ -110,7 +120,7 @@ class EasyPostObject(object):
                 cur = cur_parent
                 cur_parent = cur._parent
 
-    def __getattr__(self, k):
+    def __getattr__(self, k) -> Any:
         try:
             return self.__dict__[k]
         except KeyError:
@@ -120,36 +130,40 @@ class EasyPostObject(object):
     def __getitem__(self, k):
         return self.__dict__[k]
 
-    def get(self, k, default=None):
+    def get(self, k, default: Any = None) -> Any:
         try:
             return self[k]
         except KeyError:
             return default
 
-    def setdefault(self, k, default=None):
+    def setdefault(self, k, default: Any = None) -> Any:
         try:
             return self[k]
         except KeyError:
             self[k] = default
         return default
 
-    def __setitem__(self, k, v):
+    def __setitem__(self, k, v) -> None:
         setattr(self, k, v)
 
-    def keys(self):
+    @property
+    def keys(self) -> List[str]:
         return self._values.keys()
 
-    def values(self):
+    @property
+    def values(self) -> List[Any]:
         return self._values.keys()
 
     @classmethod
-    def construct_from(cls, values, api_key=None, parent=None, name=None):
+    def construct_from(
+        cls, values: Dict[str, Any], api_key: Optional[str] = None, parent: object = None, name: Optional[str] = None
+    ) -> object:
         """Construct an object."""
-        instance = cls(values.get("id"), api_key, parent, name)
-        instance.refresh_from(values, api_key)
+        instance = cls(easypost_id=values.get("id"), api_key=api_key, parent=parent, name=name)
+        instance.refresh_from(values=values, api_key=api_key)
         return instance
 
-    def refresh_from(self, values, api_key):
+    def refresh_from(self, values: Dict[str, Any], api_key: Optional[str] = None) -> None:
         """Update local object with changes from the API."""
         self._api_key = api_key
 
@@ -158,12 +172,12 @@ class EasyPostObject(object):
                 self.id = v
             if k in self._immutable_values:
                 continue
-            self.__dict__[k] = convert_to_easypost_object(v, api_key, self, k)
+            self.__dict__[k] = convert_to_easypost_object(response=v, api_key=api_key, parent=self, name=k)
             self._values.add(k)
             self._transient_values.discard(k)
             self._unsaved_values.discard(k)
 
-    def flatten_unsaved(self):
+    def flatten_unsaved(self) -> Dict[str, Any]:
         """Return a dict of `_unsaved_values` values from the current object."""
         values = {}
         for key in self._unsaved_values:
@@ -174,29 +188,29 @@ class EasyPostObject(object):
                 values[key] = value
         return values
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         type_string = ""
 
         if isinstance(self.get("object"), str):
             type_string = " %s" % self.get("object").encode("utf8")
 
-        json_string = json.dumps(self.to_dict(), sort_keys=True, indent=2, cls=EasyPostObjectEncoder)
+        json_string = json.dumps(obj=self.to_dict(), sort_keys=True, indent=2, cls=EasyPostObjectEncoder)
 
         return "<%s%s at %s> JSON: %s" % (type(self).__name__, type_string, hex(id(self)), json_string)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.to_json(indent=2)
 
-    def __eq__(self, other):
+    def __eq__(self, other) -> bool:
         if not isinstance(other, EasyPostObject):
             return False
         return self.__str__() == other.__str__()
 
-    def to_json(self, indent=None):
+    def to_json(self, indent: int = None) -> str:
         """Convert current object to json string."""
-        return json.dumps(self.to_dict(), sort_keys=True, indent=indent, cls=EasyPostObjectEncoder)
+        return json.dumps(obj=self.to_dict(), sort_keys=True, indent=indent, cls=EasyPostObjectEncoder)
 
-    def to_dict(self):
+    def to_dict(self) -> Dict[str, Any]:
         """Convert current object to a dict."""
 
         def _serialize(o):
@@ -215,9 +229,9 @@ class EasyPostObject(object):
 
 
 class EasyPostObjectEncoder(json.JSONEncoder):
-    def default(self, obj):
+    def default(self, obj: Any) -> Any:
         """Convert easypost object to a dict."""
         if isinstance(obj, EasyPostObject):
             return obj.to_dict()
         else:
-            return json.JSONEncoder.default(self, obj)
+            return json.JSONEncoder.default(self, o=obj)
--- a/easypost/easypost_object.py
+++ b/easypost/easypost_object.py
@@ -5,7 +5,10 @@ import easypost
 
 
 def convert_to_easypost_object(
-    response: Dict[str, Any], api_key: Optional[str] = None, parent: object = None, name: Optional[str] = None
+    response: Dict[str, Any],
+    api_key: Optional[str] = None,
+    parent: object = None,
+    name: Optional[str] = None,
 ):
     """Convert a response to an EasyPost object."""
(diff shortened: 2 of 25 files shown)


Justintime50 (changes requested):
I haven't looked at any of these yet one of the asks of the ticket was to also add `mypy` when adding type hints to ensure these were done correctly:

1. Add `mypy` (unpinned, no version constraint required) to the `DEV_REQIUREMENTS` constant in `setup.py` and install it in your project
2. Add the following to the `Makefile` (in alphabetical order with the rest of the Makefile targets:

```bash
## mypy - Run mypy type checking on the project
mypy:
	$(VIRTUAL_BIN)/mypy $(PROJECT_NAME)/ $(TEST_DIR)/
```

3. Add `mypy` (alphabetically) to the `.PHONY` line at the end of the Makefile
4. Add `mypy` to the end of the `format`, and `format-check` lines
5. Run `make mypy` and see if all these changes are correct (correct as needed) and push the changes up

Once that's done and the build is green (doing the above will get it to run on CI, no need to add an extra step since it'll be inside the `format-check` Makefile target which is already getting run) I'll take a look and stamp.

Justintime50:
@nwithan8 rebase away

jchen293:
@nwithan8  Can you add the type hint for class attributes as part of this PR? Example for Batch class.

Screen Shot 2022-03-18 at 12 55 54 PM

nwithan8:
> @nwithan8 Can you add the type hint for class attributes as part of this PR? Example for Batch class.
> 
> Screen Shot 2022-03-18 at 12 55 54 PM

I was just about to bring this up, that it seems all of our properties are programmatically generated rather than explicit. But I don't see this in master?

jchen293:
> > @nwithan8 Can you add the type hint for class attributes as part of this PR? Example for Batch class.
> > Screen Shot 2022-03-18 at 12 55 54 PM
> 
> I was just about to bring this up, that it seems all of our properties are programmatically generated rather than explicit. But I don't see this in master?

I was using the class properties from our API doc example.

nwithan8:
> > > @nwithan8 Can you add the type hint for class attributes as part of this PR? Example for Batch class.
> > > Screen Shot 2022-03-18 at 12 55 54 PM
> > 
> > 
> > I was just about to bring this up, that it seems all of our properties are programmatically generated rather than explicit. But I don't see this in master?
> 
> I was using the class properties from our API doc example.

We'll do this in a separate PR.

easypost/address.py
```diff
@@ -1,12 +1,16 @@
...
 from easypost.resource import AllResource, CreateResource
 
 
 class Address(AllResource, CreateResource):
     @classmethod
-    def create(cls, api_key=None, verify=None, verify_strict=None, **params):
-        requestor = Requestor(api_key)
+    def create(
+        cls, api_key: str = None, verify: Dict[str, Any] = None, verify_strict: Dict[str, Any] = None, **params
+    ) -> "Address":
```
Justintime50, line 12: I'm unfamiliar with the return type of a string of an object, can you point me to docs that define this? Typically I'd expect this to just be `Address` as the object itself and not a string of the word Address.

easypost/easypost_object.py
```diff
@@ -1,9 +1,10 @@
 import json
+from typing import Any, List, Optional
 
 import easypost
 
 
-def convert_to_easypost_object(response, api_key, parent=None, name=None):
+def convert_to_easypost_object(response: dict, api_key: Optional[str] = None, parent: object = None, name: str = None):
```
Justintime50, line 7: Shouldn't all of these with `None` as the fallback each be wrapped with `Optional`?

This is true throughout the changes in this PR.

easypost/easypost_object.py
```diff
@@ -62,7 +63,7 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
     }
 
     if isinstance(response, list):
-        return [convert_to_easypost_object(r, api_key, parent) for r in response]
+        return [convert_to_easypost_object(response=r, api_key=api_key, parent=parent) for r in response]
```
Justintime50, line 66: I know you didn't write it, but can we please define `r` here? It's better to be explicit than implicit and we can clean this up a bit since we are here.

easypost/easypost_object.py
```diff
@@ -73,13 +74,15 @@ def convert_to_easypost_object(response, api_key, parent=None, name=None):
...
-        return cls.construct_from(response, api_key, parent, name)
+        return cls.construct_from(values=response, api_key=api_key, parent=parent, name=name)
     else:
         return response
 
 
 class EasyPostObject(object):
-    def __init__(self, easypost_id=None, api_key=None, parent=None, name=None, **params):
+    def __init__(
+        self, easypost_id: str = None, api_key: Optional[str] = None, parent: object = None, name: str = None, **params
```
Justintime50, line 84: For these really long parameter lists, can we please use trailing commas? Black will auto format this so each is on its own line.

This is true of all throughout the repo.

easypost/easypost_object.py
```diff
@@ -135,32 +138,34 @@ def setdefault(self, k, default=None):
     def __setitem__(self, k, v):
         setattr(self, k, v)
 
-    def keys(self):
+    def keys(self) -> List:
```
Justintime50, line 141: Where possible, can we specify what it's a list of? Let's try to avoid empty `Dict` and `List` types where possible and also specify what types go in them.

easypost/easypost_object.py
```diff
@@ -180,18 +185,18 @@ def __repr__(self):
...
         if not isinstance(other, EasyPostObject):
             return False
         return self.__str__() == other.__str__()
 
-    def to_json(self, indent=None):
+    def to_json(self, indent=None) -> str:
         return json.dumps(self.to_dict(), sort_keys=True, indent=indent, cls=EasyPostObjectEncoder)
 
-    def to_dict(self):
+    def to_dict(self) -> dict:
```
Justintime50, line 199: This must be `Dict` and not `dict`, I'm not sure how this didn't fail on CI to be honest...

easypost/easypost_object.py
```diff
@@ -135,32 +138,34 @@ def setdefault(self, k, default=None):
...
             if k in self._immutable_values:
                 continue
-            self.__dict__[k] = convert_to_easypost_object(v, api_key, self, k)
+            self.__dict__[k] = convert_to_easypost_object(response=v, api_key=api_key, parent=self, name=k)
             self._values.add(k)
             self._transient_values.discard(k)
             self._unsaved_values.discard(k)
 
-    def flatten_unsaved(self):
+    def flatten_unsaved(self) -> dict:
```
Justintime50, line 168: `Dict`

easypost/easypost_object.py
```diff
@@ -135,32 +138,34 @@ def setdefault(self, k, default=None):
...
-        instance.refresh_from(values, api_key)
+    def construct_from(
+        cls, values: dict, api_key: Optional[str] = None, parent: object = None, name: str = None
+    ) -> object:
+        instance = cls(easypost_id=values.get("id"), api_key=api_key, parent=parent, name=name)
+        instance.refresh_from(values=values, api_key=api_key)
         return instance
 
-    def refresh_from(self, values, api_key):
+    def refresh_from(self, values: dict, api_key: Optional[str] = None) -> None:
```
Justintime50, line 155: `Dict`

easypost/requestor.py
```diff
@@ -11,33 +13,41 @@
 
 
 class Requestor:
-    def __init__(self, local_api_key=None):
+    def __init__(self, local_api_key: str = None):
         self._api_key = local_api_key
 
     @classmethod
-    def _objects_to_ids(cls, param):
+    def _objects_to_ids(cls, param: dict) -> dict:
```
Justintime50, line 20: `Dict`

easypost/requestor.py
```diff
@@ -11,33 +13,41 @@
...
         if isinstance(param, EasyPostObject):
             return {"id": param.id}
         elif isinstance(param, dict):
-            return {k: cls._objects_to_ids(v) for k, v in param.items()}
-        elif isinstance(param, list):
-            return [cls._objects_to_ids(v) for v in param]
+            data = {}
```
Justintime50, line 24: With these changes here, have we tried to force code to run through them by making a fresh API call (no VCR)?

easypost/requestor.py
```diff
@@ -11,33 +13,41 @@
...
+                    data[k] = [cls._objects_to_ids(item) for item in v]  # type: ignore
+                else:
+                    data[k] = cls._objects_to_ids(v)  # type: ignore
+            return data
         else:
             return param
 
-    def request(self, method, url, params=None, api_key_required=True):
+    def request(
+        self, method: str, url: str, params: dict = None, api_key_required: bool = True
```
Justintime50, line 35: `Dict`

easypost/requestor.py
```diff
@@ -11,33 +13,41 @@
...
+    ) -> Tuple[dict, Optional[str]]:
```
Justintime50, line 36: `Dict` (there are a ton of these, I'm going to stop explicitly marking them, please check this PR for any that need changing.

easypost/requestor.py
```diff
@@ -48,24 +58,18 @@ def request_raw(self, method, url, params=None, api_key_required=True):
...
             "client_version": VERSION,
             "lang": "python",
             "publisher": "easypost",
             "request_lib": request_lib,
+            "lang_version": platform.python_version(),
+            "platform": platform.platform(),
+            "uname": " ".join(platform.uname()),
+            "implementation": platform.python_implementation(),
         }
-        for attr, func in (
```
Justintime50, line 59: Why was this refactored here? By doing so we lose the try/except that may have been needed in some situations

easypost/shipment.py
```diff
@@ -1,65 +1,59 @@
...
+        self.refresh_from(values=response, api_key=api_key)
         return self
 
-    def lowest_rate(self, carriers=None, services=None):
+    def lowest_rate(self, carriers: List[str] = None, services: List[str] = None) -> Rate:
         carriers = carriers or []
         services = services or []
         lowest_rate = None
 
-        try:
```
Justintime50, line 53: Hmm... agree that the removal of this is a good thing because it's supposed to be a list, but it's still a breaking change that should get called out in the CHANGELOG in case someone was relying on this.

easypost/user.py
```diff
@@ -1,60 +1,54 @@
...
+        response, api_key = requestor.request(method="get", url="/api_keys")
+        return util.convert_to_easypost_object(response=response, api_key=api_key)
 
-    def api_keys(self):
+    def api_keys(self) -> Union[Union[Callable[[], List], List[Any]], Any]:
         api_keys = self.all_api_keys()
 
         if api_keys.id == self.id:
-            my_api_keys = api_keys.keys
+            return api_keys.keys
```
Justintime50, line 44: Are we returning 3 times here because of `mypy` being weird? We typically try to avoid returning multiple times and return once at the end.

easypost/version.py
```diff
@@ -1,6 +1,8 @@
 VERSION = "6.0.0"
 
-if "-" in VERSION:
-    VERSION_INFO = tuple([int(v) for v in VERSION.split("-")[0].split(".")] + VERSION.split("-")[1:])
-else:
-    VERSION_INFO = tuple(int(v) for v in VERSION.split("."))
+elems = VERSION.split("-")
+nums = [str(v) for v in elems[0].split(".")]
```
Justintime50, line 4: Explicit over implicit - by doing so the reader (especially while skimming) has to do much less mental work to determine what was meant here and filter our any possibilities of what it could have meant.

```suggestion
elements = VERSION.split("-")
numbers = [str(v) for v in elems[0].split(".")]
```

Justintime50 (dismissed):

Justintime50 (commented):
We'll also need to add an empty `py.typed` file into the `easypost` directory and then add `package_data={'easypost': ['py.typed']},` to `setup.py` so that users will actually have our type hints available to them in their own apps.

easypost/address.py
nwithan8: You can type hint a class by name if it's inside of itself or not imported: https://peps.python.org/pep-0484/#forward-references
Otherwise, the type hint would throw an error saying it can't find the referred class.

easypost/user.py
nwithan8: Yes

easypost/address.py
Justintime50: Ah very good, thanks!

easypost/requestor.py
nwithan8: Getting some errors, but seem to be unrelated to this change.

easypost/requestor.py
nwithan8: Reverted in 94d6bd1b26f8d68dcb940fd8f3f37d758f270568

easypost/shipment.py
nwithan8: Resolved in 94d6bd1b26f8d68dcb940fd8f3f37d758f270568

Justintime50:
@nwithan8 lint needs to be fixed.

nwithan8:
> @nwithan8 lint needs to be fixed.

It's related to this: https://github.com/EasyPost/easypost-python/pull/163#discussion_r830366136
`mypy` doesn't like the fact that we're just executing a function, with no concept of what said function will return.
We'll need to decide how to handle this.

easypost/requestor.py
Justintime50: > > @nwithan8 lint needs to be fixed.
> 
> It's related to this: [#163 (comment)](https://github.com/EasyPost/easypost-python/pull/163#discussion_r830366136) `mypy` doesn't like the fact that we're just executing a function, with no concept of what said function will return. We'll need to decide how to handle this.

Let's get the build passing in the most compatible way to what was there unless we can prove the cold is old and unnecessary anymore.

nwithan8: Resolved in [f20ed30](https://github.com/EasyPost/easypost-python/pull/163/commits/f20ed308b210f3a207f8c94fa1eac9b6ef2ceefe)

easypost/requestor.py
```diff
@@ -131,21 +149,23 @@ def requests_request(self, method, abs_url, headers, params):
...
-            fetch_args["url"] = self.add_params_to_url(abs_url, params)
         else:
             raise Error(f"Bug discovered: invalid request method: {method}. Please report to {SUPPORT_EMAIL}.")
 
         try:
-            from google.appengine.api import urlfetch
+            from google.appengine.api import urlfetch  # type: ignore
 
-            result = urlfetch.fetch(**fetch_args)
+            result = urlfetch.fetch(**fetch_args)  # type: ignore
```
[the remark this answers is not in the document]
nwithan8: The `#type: ignore` is an indicator to `mypy` to ignore type-hinting issues here. Otherwise, this line is a nightmare because it could realistically be several different types and that spreads up the chain to the rest of the app.

tests/conftest.py
```diff
@@ -13,11 +13,11 @@ def pytest_sessionstart(session):
...
     try:
-        from google.appengine.ext import testbed
+        from google.appengine.ext import testbed  # type: ignore
 
-        session.appengine_testbed = testbed.Testbed()
-        session.appengine_testbed.activate()
-        session.appengine_testbed.init_urlfetch_stub()
+        session.appengine_testbed = testbed.Testbed()  # type: ignore
+        session.appengine_testbed.activate()  # type: ignore
+        session.appengine_testbed.init_urlfetch_stub()  # type: ignore
```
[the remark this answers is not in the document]
nwithan8: The `#type: ignore` is an indicator to `mypy` to ignore type-hinting issues here. Otherwise, this line is a nightmare because it could realistically be several different types and that spreads up the chain to the rest of the app.

jchen293 (dismissed):

easypost/easypost_object.py
```diff
@@ -1,9 +1,12 @@
 import json
+from typing import Any, Dict, List, Optional
 
 import easypost
 
 
-def convert_to_easypost_object(response, api_key, parent=None, name=None):
+def convert_to_easypost_object(
+    response: Dict[str, Any], api_key: Optional[str] = None, parent: object = None, name: Optional[str] = None
```
Justintime50, line 8: Nit: trailing comma to have Black break this up to separate lines.

easypost/easypost_object.py
```diff
@@ -120,36 +130,40 @@ def __getattr__(self, k):
...
 
-    def values(self):
+    @property
+    def values(self) -> List[Any]:
         return self._values.keys()
 
     @classmethod
-    def construct_from(cls, values, api_key=None, parent=None, name=None):
+    def construct_from(
+        cls, values: Dict[str, Any], api_key: Optional[str] = None, parent: object = None, name: Optional[str] = None
```
Justintime50, line 159: Nit: Trailing comma

easypost/requestor.py
```diff
@@ -11,37 +13,45 @@
...
+                    data[k] = [cls._objects_to_ids(item) for item in v]  # type: ignore
+                else:
+                    data[k] = cls._objects_to_ids(v)  # type: ignore
+            return data
         else:
             return param
 
-    def request(self, method, url, params=None, api_key_required=True):
+    def request(
+        self, method: str, url: str, params: Optional[Dict[str, Any]] = None, api_key_required: bool = True
```
Justintime50, line 37: Nit: Trailing comma

easypost/requestor.py
```diff
@@ -83,29 +93,37 @@ def request_raw(self, method, url, params=None, api_key_required=True):
...
-    def interpret_response(self, http_body, http_status):
+    def interpret_response(self, http_body: str, http_status: int) -> Dict[str, Any]:
         """Interpret the response body we receive from the API."""
         try:
             response = json.loads(http_body)
-        except Exception:
-            raise Error("Invalid response body from API: (%d) %s " % (http_status, http_body), http_status, http_body)
+        except JSONDecodeError:
+            raise Exception(
+                "Unable to parse response body from API: %s\nHTTP response code was: %d" % (http_body, http_status)
```
Justintime50, line 114: f-string please

easypost/requestor.py
```diff
@@ -11,37 +13,45 @@
...
-        response = self.interpret_response(http_body, http_status)
+        http_body, http_status, my_api_key = self.request_raw(
+            method=method, url=url, params=params, api_key_required=api_key_required
+        )
+        response = self.interpret_response(http_body=http_body, http_status=http_status)
         return response, my_api_key
 
-    def request_raw(self, method, url, params=None, api_key_required=True):
+    def request_raw(
+        self, method: str, url: str, params: Optional[Dict[str, Any]] = None, api_key_required: bool = True
```
Justintime50, line 49: Nit: Trailing comma

easypost/requestor.py
```diff
@@ -83,29 +93,37 @@ def request_raw(self, method, url, params=None, api_key_required=True):
...
+                "Unable to parse response body from API: %s\nHTTP response code was: %d" % (http_body, http_status)
+            )
         if not (200 <= http_status < 300):
-            self.handle_api_error(http_status, http_body, response)
+            self.handle_api_error(http_status=http_status, http_body=http_body, response=response)
         return response
 
-    def requests_request(self, method, abs_url, headers, params):
+    def requests_request(
+        self, method: str, abs_url: str, headers: Dict[str, Any], params: Dict[str, Any]
```
Justintime50, line 121: Nit: Trailing comma

easypost/requestor.py
```diff
@@ -131,21 +149,23 @@ def requests_request(self, method, abs_url, headers, params):
             )
         return http_body, http_status
 
-    def urlfetch_request(self, method, abs_url, headers, params):
+    def urlfetch_request(
+        self, method: str, abs_url: str, headers: Dict[str, Any], params: Dict[str, Any]
```
Justintime50, line 153: Nit: Trailing comma

easypost/requestor.py
```diff
@@ -182,11 +202,11 @@ def encode_url_params(self, params):
             elif isinstance(value, datetime.datetime):
                 value = int(time.mktime(value.timetuple()))  # to UTC timestamp
             converted_params.append((key, value))
-            return urlencode(converted_params)
```
Justintime50, line 185: How for so many years was this incorrectly indented and still worked?

easypost/user.py
```diff
@@ -1,66 +1,62 @@
...
         else:
             for child in api_keys.children:
                 if child.id == self.id:
-                    my_api_keys = child.keys
-                    break
+                    return child.keys
 
-        return my_api_keys
+        return []
```
Justintime50, line 56: Why is this here? There shouldn't ever be a time where we return an empty list as either the request itself will fail or a list of API keys will get returned. You can't make a request to get API keys with an API key and not get one back.

setup.py
```diff
@@ -11,9 +12,15 @@
     "pytest-cov==3.*",
     "pytest-vcr==1.*",
     "pytest==7.*",
+    "types-requests",
+    "types-urllib3",
     "vcrpy==4.*",
 ]
 
+CPYTHON_DEV_REQUIREMENTS = [
```
Justintime50, line 20: I hate this. At the earliest convenience (once we can confirm nobody is using `pypy`, let's remove this weirdness.

Can you leave an inline comment as to why these are broken out the way they are?

Justintime50 (changes requested):

easypost/requestor.py
nwithan8: ¯\_(ツ)_/¯

easypost/user.py
nwithan8: Then it should never reach this line 😉 , but if it ever does.... but it should never, but just in case...

easypost/requestor.py
nwithan8: It wasn't until we split them into new files just a few weeks ago: https://github.com/EasyPost/easypost-python/blob/v6.0.0/easypost/__init__.py#L179

easypost/user.py
Justintime50: It can literally never get here haha. Code that serves no purpose shouldn't be in the repo.

Justintime50 (approved):

easypost/user.py
nwithan8: Watch. One day something catastrophic will happen that will trigger this, and rather than the user getting nothing back from this function, and attempting to use a `None` later that will blow up their code stack, they'll get just a friendly empty list instead.

nwithan8: In any other programming language, a function written without this would complain that not all routes return a value, and may not even compile. Just because Python doesn't care doesn't mean we shouldn't.

nwithan8: See, `mypy` knows :)
image

jchen293 (approved):

PR #165, opened 2022-03-18
fix: import util.convert_to_easypost_object directly

Justintime50:
There is no need to wrap `convert_to_easypost_object` in `util` as it provides no value and is confusing; instead, we should just import the function directly.


jchen293 (approved):
Ah nice, how did I miss this in the overhaul PR.

nwithan8 (approved):

PR #166, opened 2022-03-18
collect python_implementation in easypost-python user-agent

Roguelazer:
This should be backported to all previous series (3.x, 4.x, 5.x, etc).

### DIFF
--- a/easypost/requestor.py
+++ b/easypost/requestor.py
@@ -60,6 +60,7 @@ class Requestor:
             ("lang_version", platform.python_version),
             ("platform", platform.platform),
             ("uname", lambda: " ".join(platform.uname())),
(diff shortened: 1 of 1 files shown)


nwithan8:
That logic changed in syntax in another PR. I'll add this change into that PR to avoid merge conflicts.

Roguelazer:
Like I said, this has to be backported and new releases of existing stable series need to be done, so it may not make sense to fold this into some future-work PR.

jontsai (approved):

Justintime50 (approved):

nwithan8 (approved):

jchen293 (approved):

PR #167, opened 2022-03-21
Support for `columns` and `additional_columns` on Report creation

jchen293:
This PR adds support for specifying `columns` and `additional_columns` in the params for the report create method.

Adds two unit tests for this feature. E2E test and download the CSV report to check if the result is correct.

Additional columns and columns in the param, total three columns (from_name, from_company, usps_zone)
Screen Shot 2022-03-21 at 3 10 50 PM

Only columns in the param, only one column (usps_zone)
Screen Shot 2022-03-21 at 3 12 34 PM


easypost/report.py
```diff
@@ -7,7 +7,20 @@ class Report(AllResource, CreateResource):
     @classmethod
     def create(cls, api_key=None, **params):
         requestor = Requestor(api_key)
-        url = "%s/%s" % (cls.class_url(), params["type"])
+        url = "%s/%s" % (cls.class_url(), params["type"]) + "?"
+
+        if "columns" in params:
+            for column in params.get("columns", []):
+                url += "columns[]=" + str(column) + "&"
```
nwithan8, line 14: Nit: Since we only support 3.6+ now, we can start using f-strings

tests/test_report.py
```diff
@@ -11,6 +11,32 @@ def create_report(report_type, report_start_date, report_end_date):
...
+def test_report_create_with_columns(report_start_date, report_end_date):
+    report = easypost.Report.create(
+        type="shipment",
+        start_date=report_start_date,
+        end_date=report_end_date,
+        columns=["usps_zone"],
+    )
+
+    assert isinstance(report, easypost.Report)
+    assert str.startswith(report.id, "shprep_")
```
nwithan8, line 37: We don't need the ID check

tests/test_report.py
```diff
@@ -11,6 +11,32 @@ def create_report(report_type, report_start_date, report_end_date):
...
+def test_report_create_with_additional_columns(report_start_date, report_end_date):
+    report = easypost.Report.create(
+        type="shipment",
+        start_date=report_start_date,
+        end_date=report_end_date,
+        additional_columns=["from_name", "from_company"],
+    )
+
+    assert isinstance(report, easypost.Report)
+    assert str.startswith(report.id, "shprep_")
```
nwithan8, line 24: We don't need the ID check.
Please add a comment explaining why we're not doing any more tests (something like `Reports are queued, so we can't wait for completion. Verifying columns would require parsing CSV. Verify correct parameters via URL in cassette`)

nwithan8 (changes requested):

tests/test_report.py
jchen293: I believe from the discussion yesterday, we are going to have these two assertions, similar to other unit tests for `report`.

easypost/report.py
jchen293: good suggestion

tests/test_report.py
jchen293: added

tests/test_report.py
Justintime50: Not exactly, we don't need to assert the ID since all we are testing is if the report gets created correctly. We are testing the ID elsewhere. We should assert that what comes back is a report (indicating success with the params passed) and then leaving a comment similar to what we do with batches saying we can't test this more in-depth because the report gets queued for generation and most likely isn't available quick enough for us to retrieve.

jchen293: fixed.

nwithan8 (approved):

Justintime50:
@jchen293 please rebase this against master before I take a look.

jchen293:
> @jchen293 please rebase this against master before I take a look.

rebased

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         """Create a report."""
         requestor = Requestor(api_key)
-        url = "%s/%s" % (cls.class_url(), params["type"])
+        url = "%s/%s" % (cls.class_url(), params["type"]) + "?"
```
Justintime50, line 11: Please us an f string here, it's much easier to read than mentally trying to slot in variables to a placeholder.

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
...
+
+        if "columns" in params:
+            for column in params.get("columns", []):
+                url += f"columns[]={str(column)}&"
```
Justintime50, line 15: I'm unsure that we need to wrap this in `str()` since we expect them to be strings already. If they passed in say integers, then coercing it to a string won't do any good because it will still fail at the API level.

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
...
+            # Delete the columns from the param sinces it's added in the url query.
```
Justintime50, line 16:
```suggestion
            # Delete the columns from the params since it's added in the url query.
```

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
...
+            params.pop("columns")
+
+        if "additional_columns" in params:
+            for column in params.get("additional_columns", []):
+                url += f"additional_columns[]={str(column)}&"
+            # Delete the additional columns from the param sinces it's added in the url query.
```
Justintime50, line 22:
```suggestion
            # Delete the additional columns from the params since it's added in the url query.
```

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
...
+
+        if "columns" in params:
+            for column in params.get("columns", []):
+                url += f"columns[]={str(column)}&"
+            # Delete the columns from the param sinces it's added in the url query.
+            params.pop("columns")
+
+        if "additional_columns" in params:
+            for column in params.get("additional_columns", []):
+                url += f"additional_columns[]={str(column)}&"
```
Justintime50, line 21: I'm unsure that we need to wrap this in str() since we expect them to be strings already. If they passed in say integers, then coercing it to a string won't do any good because it will still fail at the API level.

Justintime50 (changes requested):

easypost/report.py
jchen293: you're correct, this is redundant

easypost/report.py
jchen293: you're correct, this is redundant

jchen293: you're correct, this is redundant

nwithan8 (dismissed):
I imagine f-strings will be the next thing to tackle in this library.
Looks good.

easypost/report.py
```diff
@@ -8,7 +8,22 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         """Create a report."""
         requestor = Requestor(api_key)
-        url = "%s/%s" % (cls.class_url(), params["type"])
+        class_url = cls.class_url()
```
Justintime50, line 11: Assigning intermediate variables is good in two scenarios:

1. When the value you are assigning to a variable isn't immediately clear what it's for/doing
2. When that same value will get used multiple times

Since 1 isn't true as the name of the variable is directly in the name of the value and 2 isn't true because it's only used once, we can remove this intermediate variable and clean this code up a bit by simply calling `cls.class_url()` directly in the `url` variable below.

easypost/report.py
```diff
@@ -8,7 +8,22 @@ class Report(AllResource, CreateResource):
...
+        type = params["type"]
```
Justintime50, line 12: Please see my comment above about intermediate variables - the same applies for this `type` variable here.

easypost/report.py
```diff
@@ -8,7 +8,22 @@ class Report(AllResource, CreateResource):
...
+        url = f"{class_url}/{type}?"
+
+        if "columns" in params:
```
Justintime50, line 15: This would be safer as `if params.get('columns')`

easypost/report.py
```diff
@@ -8,7 +8,22 @@ class Report(AllResource, CreateResource):
...
+            for column in params.get("columns", []):
+                url += f"columns[]={column}&"
+            # Delete the columns from the params since it's added in the url query.
+            params.pop("columns")
+
+        if "additional_columns" in params:
```
Justintime50, line 21: This would be safer as `if params.get('additional_columns')`

Justintime50 (changes requested):

nwithan8 (dismissed):

easypost/report.py
```diff
@@ -8,7 +8,20 @@ class Report(AllResource, CreateResource):
     def create(cls, api_key=None, **params):
         """Create a report."""
         requestor = Requestor(api_key)
-        url = "%s/%s" % (cls.class_url(), params["type"])
+        url = f"{cls.class_url()}/{params.get('type')}?"
```
Justintime50, line 11: If the user doesn't specify a type, this will now populate the url with `None` and will fail at the API level, which is probably fine because previously we'd throw a `KeyError` if there was no type. 

In a follow up, we should probably add validation for reports across all libs to check if the type is present and error without a type specified.

Justintime50 (approved):

nwithan8 (approved):

easypost/report.py
jchen293: Sounds good, I will add an internal ticket for this.

PR #172, opened 2022-04-04
add conditional check if the http is 204 to prevent error being thrown.

jchen293:
When deleting a child user, our API endpoint will return HTTP 204 with an empty response body. This is a special case because HTTP 204 means the request is succeeded but doesn't have a response body, in our Python library it will throw an error because we never handle this corner, the root of the cause is `json.loads` in the `interpret_response` method because an error will be thrown when the response body is empty even thought the HTTP request is succeeded. See the screenshot below.

This PR checks if the HTTP Status response is 204, since the response body is empty the `json.loads()` will throw an error. Adding an if statement to handle the corner case.

Error stack: 
Screen Shot 2022-04-04 at 1 51 13 PM

### DIFF
--- a/easypost/requestor.py
+++ b/easypost/requestor.py
@@ -114,6 +114,9 @@ class Requestor:
 
     def interpret_response(self, http_body: str, http_status: int) -> Dict[str, Any]:
         """Interpret the response body we receive from the API."""
+        if http_status == 204:
+            # HTTP 204 does not have any response body and we can just return here
+            return {}
         try:
             response = json.loads(http_body)
         except JSONDecodeError:
--- a/tests/cassettes/test_carrier_account_delete.yaml
+++ b/tests/cassettes/test_carrier_account_delete.yaml
@@ -25,11 +25,11 @@ interactions:
   response:
     body:
       string: !!binary |
-        H4sIAAAAAAAAA4SQwWrDMBBEfyXo7ICt2KjxLfReAmkvvZjVal1UtpKR5IAJ/vdKmDbkFHTa2bcz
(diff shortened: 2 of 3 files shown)


Justintime50:
Can you please elaborate on why this is needed and provide the error stack that occurs when this fix is not in place?

jchen293:
> Can you please elaborate on why this is needed and provide the error stack that occurs when this fix is not in place?

I have elaborate the PR description.

Justintime50:
We should re-record the cassette that will use this updated logic as part of this PR to ensure it works.

jchen293:
> We should re-record the cassette that will use this updated logic as part of this PR to ensure it works.

`Webhook` and `carrier_account` are the only two that use this updated delete logic. For deleting a child user, I have E2E locally tested.

Justintime50 (approved):

Justintime50:
I'm curious why this didn't break before?

jchen293:
> I'm curious why this didn't break before?

`test_user_delete` is skipped in our unit test and this is by far the only API endpoint that returns HTTP 204 with an empty body.

nwithan8:
As opposed to specifically checking for a 204 status code, why don't we handle an empty JSON body inside `interpret_response`?

jchen293:
> As opposed to specifically checking for a 204 status code, why don't we handle an empty JSON body inside `interpret_response`?

HTTP 204 guarantees the return JSON body is empty, I am unsure if there will be a case from the API endpoint returning an empty JSON body for HTTP code that is other than 204.

Justintime50 (approved):

nwithan8 (approved):

PR #175, opened 2022-04-07
remove message conditional check in address create and verify.

jchen293:
- Removes dead message conditional check in Address create and verify
- Updates a few address assertions to match internal API changes


Justintime50 (changes requested):
Please also remove the `carrier` logic from this function (I apologize, it's probably not in the original ticket) but is needed at this time too.

Justintime50 (approved):

Justintime50 (commented):
Can we also add what was done here to the CHANGELOG (removed carrier param, remove conditional message check)?

Justintime50 (approved):

jontsai (approved):

PR #185, opened 2022-04-26
feat: adds lowest_rate to more objects

Justintime50:
# Description

- Adds `lowest_rate` functions to Orders and Pickups
- Adds a `Shipment.get_lowest_smartrate` function and a `shipment.lowest_smartrate()` function

# Testing

- [x] Added various new unit tests to ensure all `lowest_rate` functions work

# Pull Request Type

### DIFF
--- a/easypost/order.py
+++ b/easypost/order.py
@@ -26,7 +26,7 @@ class Order(CreateResource):
         return self
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
-        """Get the lowest rate of a order."""
-        lowest_rate = Util.get_lowest_object_rate(self, carriers, services)
+        """Get the lowest rate of an order."""
+        lowest_rate = Util._get_lowest_object_rate(self, carriers, services)
 
         return lowest_rate
--- a/easypost/order.py
+++ b/easypost/order.py
@@ -5,7 +5,7 @@ from easypost.requestor import (
     Requestor,
 )
 from easypost.resource import CreateResource
-from easypost.util import Util
+from easypost.util import get_lowest_object_rate
 
 
 class Order(CreateResource):
@@ -27,6 +27,6 @@ class Order(CreateResource):
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
         """Get the lowest rate of an order."""
-        lowest_rate = Util._get_lowest_object_rate(self, carriers, services)
+        lowest_rate = get_lowest_object_rate(self, carriers, services)
 
         return lowest_rate
--- a/easypost/order.py
+++ b/easypost/order.py
@@ -26,7 +26,7 @@ class Order(CreateResource):
         return self
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
-        """Get the lowest rate of an order."""
+        """Get the lowest rate of this order."""
         lowest_rate = get_lowest_object_rate(self, carriers, services)
 
         return lowest_rate
--- a/easypost/order.py
+++ b/easypost/order.py
@@ -1,8 +1,11 @@
+from typing import List
+
 from easypost.requestor import (
     RequestMethod,
     Requestor,
 )
 from easypost.resource import CreateResource
+from easypost.util import Util
 
 
 class Order(CreateResource):
@@ -21,3 +24,9 @@ class Order(CreateResource):
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
+        """Get the lowest rate of a order."""
+        lowest_rate = Util.get_lowest_object_rate(self, carriers, services)
+
+        return lowest_rate
--- a/easypost/pickup.py
+++ b/easypost/pickup.py
@@ -26,7 +26,7 @@ class Pickup(CreateResource):
         return self
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
-        """Get the lowest rate of a pickup."""
+        """Get the lowest rate of this pickup."""
         lowest_rate = get_lowest_object_rate(self, carriers, services, "pickup_rates")
 
         return lowest_rate
--- a/easypost/pickup.py
+++ b/easypost/pickup.py
@@ -27,6 +27,6 @@ class Pickup(CreateResource):
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
         """Get the lowest rate of a pickup."""
-        lowest_rate = Util.get_lowest_object_rate(self, carriers, services, "pickup_rates")
+        lowest_rate = Util._get_lowest_object_rate(self, carriers, services, "pickup_rates")
 
         return lowest_rate
--- a/easypost/pickup.py
+++ b/easypost/pickup.py
@@ -5,7 +5,7 @@ from easypost.requestor import (
     Requestor,
 )
 from easypost.resource import CreateResource
-from easypost.util import Util
+from easypost.util import get_lowest_object_rate
 
 
 class Pickup(CreateResource):
@@ -27,6 +27,6 @@ class Pickup(CreateResource):
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
         """Get the lowest rate of a pickup."""
-        lowest_rate = Util._get_lowest_object_rate(self, carriers, services, "pickup_rates")
+        lowest_rate = get_lowest_object_rate(self, carriers, services, "pickup_rates")
 
         return lowest_rate
--- a/easypost/pickup.py
+++ b/easypost/pickup.py
@@ -1,8 +1,11 @@
+from typing import List
+
 from easypost.requestor import (
     RequestMethod,
     Requestor,
 )
 from easypost.resource import CreateResource
+from easypost.util import Util
 
 
 class Pickup(CreateResource):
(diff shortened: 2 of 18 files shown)


easypost/util.py
```diff
@@ -0,0 +1,36 @@
+from typing import List
+
+from easypost.easypost_object import EasyPostObject
+from easypost.error import Error
+
+
+class Util:
```
nwithan8, line 7: nit: we don't need a Util class. This can just be an assortment of functions in a `util.py` file that we import individually as needed. The way it's configured now, every function in `Util` will be available everywhere we import `Util`

easypost/util.py
```diff
@@ -0,0 +1,36 @@
...
+    @staticmethod
+    def _get_lowest_object_rate(
+        easypost_object: EasyPostObject,
+        carriers: List[str] = None,
+        services: List[str] = None,
+        rates_key: str = "rates",
+    ):
```
nwithan8: Add a return type hint, as well as a docstring (with parameters) for this method

easypost/util.py
```diff
@@ -0,0 +1,36 @@
...
+        """Gets the lowest rate of an EasyPost object such as a Shipment, Order, or Pickup."""
+        carriers = carriers or []
+        services = services or []
+        lowest_rate = None
+
+        carriers = [carrier.lower() for carrier in carriers]
+        services = [service.lower() for service in services]
+
+        for rate in easypost_object[rates_key]:
+            if len(carriers) > 0 and rate.carrier.lower() not in carriers:
```
nwithan8: nit: could be more Pythonic and just say `if carriers`, since an empty list == False in Python. Same with `services` below.

easypost/util.py
```diff
@@ -0,0 +1,36 @@
...
+    ):
+        """Gets the lowest rate of an EasyPost object such as a Shipment, Order, or Pickup."""
+        carriers = carriers or []
+        services = services or []
+        lowest_rate = None
+
+        carriers = [carrier.lower() for carrier in carriers]
+        services = [service.lower() for service in services]
+
+        for rate in easypost_object[rates_key]:
```
nwithan8: `easypost_object[rates_key]` could produce `None` if it doesn't exist (rates haven't been loaded in the object yet), which will produce an error. Try `easypost_object.get(rates_key, [])` to default to an empty list if `rates_key` doesn't exist.

easypost/pickup.py
```diff
@@ -21,3 +24,9 @@ def cancel(self, **params) -> "Pickup":
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
+        """Get the lowest rate of a pickup."""
```
nwithan8, line 29: "this pickup", not "a pickup", since it's dealing with a specific instance.

easypost/order.py
```diff
@@ -21,3 +24,9 @@ def buy(self, **params) -> "Order":
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    def lowest_rate(self, carriers: List[str] = None, services: List[str] = None):
```
nwithan8, line 28: "this order", not "an order", since it's dealing with a specific instance.

easypost/shipment.py
```diff
@@ -62,24 +66,46 @@ def label(self, **params) -> "Shipment":
 
     def lowest_rate(self, carriers: List[str] = None, services: List[str] = None) -> Rate:
         """Get the lowest rate of a shipment."""
```
nwithan8, line 68: "this shipment", not "a shipment", since it's dealing with a specific instance.

easypost/shipment.py
```diff
@@ -62,24 +66,46 @@ def label(self, **params) -> "Shipment":
...
-        services = services or []
-        lowest_rate = None
+        lowest_rate = Util._get_lowest_object_rate(self, carriers, services)
+
+        return lowest_rate
 
-        carriers = [c.lower() for c in carriers]
-        services = [service.lower() for service in services]
+    def lowest_smartrate(self, delivery_days: Optional[int] = None, delivery_accuracy: Optional[str] = None) -> Rate:
+        """Get the lowest smartrate of a shipment."""
```
nwithan8, line 74: "this shipment", not "a shipment", since it's dealing with a specific instance.

easypost/shipment.py
```diff
@@ -62,24 +66,46 @@ def label(self, **params) -> "Shipment":
...
+        smartrates = self.get_smartrates()
+        lowest_smartrate = self.get_lowest_smartrate(smartrates, delivery_days, delivery_accuracy)
```
nwithan8, line 76: `get_lowest_smartrate` can be a static function (no need for `self.`) since it doesn't use anything from this particular instance of Shipment

easypost/util.py
Justintime50: Good point. Flattened the module.

easypost/shipment.py
```diff
@@ -62,24 +66,46 @@ def label(self, **params) -> "Shipment":
...
 
-            if lowest_rate is None or float(rate.rate) < float(lowest_rate.rate):
-                lowest_rate = rate
+        for rate in smartrates:
+            if delivery_days and delivery_accuracy:
+                if delivery_accuracy not in valid_delivery_accuracy_values:
+                    raise Error(
+                        message=f"Invalid delivery_accuracy value, must be one of: {valid_delivery_accuracy_values}"
+                    )
+                elif rate["time_in_transit"][delivery_accuracy] > delivery_days:
```
nwithan8, line 102: What happens if the accuracy is not greater than delivery_days?
Also, greater than, or greater than or equal to?

nwithan8 (changes requested):

easypost/shipment.py
```diff
@@ -62,24 +66,46 @@ def label(self, **params) -> "Shipment":
...
+        lowest_smartrate = None
 
-            if lowest_rate is None or float(rate.rate) < float(lowest_rate.rate):
-                lowest_rate = rate
+        for rate in smartrates:
+            if delivery_days and delivery_accuracy:
+                if delivery_accuracy not in valid_delivery_accuracy_values:
+                    raise Error(
+                        message=f"Invalid delivery_accuracy value, must be one of: {valid_delivery_accuracy_values}"
+                    )
```
jchen293, line 101: I would move these two if statements before the loop since we don't need to check them every time inside the loop.

easypost/shipment.py
jchen293: I think the current code should be fine as we are only skipping the `delivery_date` that is greater than the `delivery_accuracy`. Line 105-106 will update the `lowest_rate` if the rate is cheaper or never assigned.

easypost/util.py
Justintime50: True. This was just copied over from the original logic to a different file. Will simplify it.

easypost/util.py
Justintime50: Although true, as a point of clarification, a rates key should always exist on these object. But this is a good best practice regardless.

easypost/shipment.py
Justintime50: `get_lowest_smartrate` is already static? What do you mean here?

easypost/shipment.py
Justintime50: Correct, the code should be what we want as is because we only want to skip making it the lowest rate if it doesn't meet the criteria (if it'll take too long to be delivered). If the `delivery_days` requested is less than or equal to the accuracy, we'll make that the lowest rate until we get to the lowest of them all and we return that.

easypost/util.py
Justintime50: I've tried here and unfortunately to import `Rate` (the return type), we get a circular dependency error at runtime.

easypost/shipment.py
jchen293: There is a `self` in the first parameter.

Justintime50: Not in `get_lowest_smartrate`. You guys are probably referencing `lowest_smartrate` which definitely needs self because you call it on the Shipment object.

easypost/util.py
nwithan8: You can use a [forward reference](https://peps.python.org/pep-0484/#forward-references), i.e. `) -> 'Rate':`

easypost/shipment.py
nwithan8: You call it here as `self.get_lowest_smartrate`

Justintime50: Would you like me to call it with `Shipment.get_lowest_smartrate`? We need `self` for the line directly above it because we need to `get_smartrates` of the shipment object before we can get the lowest smartrate, so `self` will already be in use in this function and is required earlier.

easypost/util.py
Justintime50: That doesn't work either unfortunately because `"Rate"` not defined, linting complains without `Rate` being imported which is what leads to the circular import error.

easypost/shipment.py
```diff
@@ -61,25 +65,46 @@ def label(self, **params) -> "Shipment":
...
+            "percentile_95",
+            "percentile_97",
+            "percentile_99",
+        }
+        lowest_smartrate = None
 
-            if lowest_rate is None or float(rate.rate) < float(lowest_rate.rate):
-                lowest_rate = rate
+        if delivery_accuracy and delivery_accuracy not in valid_delivery_accuracy_values:
+            raise Error(message=f"Invalid delivery_accuracy value, must be one of: {valid_delivery_accuracy_values}")
```
jchen293, line 97: Should an error also be thrown if the user does not pass in `delivery_days` or `delivery_accuracy`?

Justintime50: No. Sometimes users may just want the lowest smartrate. Though smartrates are just rates and they could get the lowest rate normally if they wanted to use no filters. What I'll do instead is make the parameters required for the smartrate functions.

nwithan8 (dismissed):

jchen293 (approved):

jchen293 (approved):

nwithan8 (approved):

PR #196, opened 2022-05-26
feat: adds user billing functions

Justintime50:
# Description

Adds `PaymentMethod.all()`, `CreditCard.fund()`, and `CreditCard.delete()` functions

# Testing

- [x] Added unit test for retrieve all payment methods
- [x] E2E tested (as much as possible) that deleting a credit card and funding a wallet would at least hit the right endpoints, can't test these functions due to not having a valid CC to use

# Pull Request Type

### DIFF
--- a/easypost/credit_card.py
+++ b/easypost/credit_card.py
@@ -26,15 +26,10 @@ class CreditCard(CreateResource, Resource):
 
         primary_or_secondary = params.get("primary_or_secondary")
         payment_method_to_use = payment_method_map.get(primary_or_secondary) if primary_or_secondary else None
+        card_id = payment_methods[payment_method_to_use]["id"] if payment_method_to_use else None
 
-        if (
-            payment_method_to_use is None
-            or payment_methods[payment_method_to_use] is None
-            or "card_" not in payment_methods[payment_method_to_use].get("id")
-        ):
+        if payment_method_to_use is None or card_id is None or "card_" not in card_id:
             raise Error(message="The chosen payment method is not a credit card. Please try again.")
-
-        card_id = payment_method_to_use.get("id")
 
         requestor = Requestor(local_api_key=api_key)
         url = f"{cls.class_url()}/{card_id}/charges"
--- a/easypost/credit_card.py
+++ b/easypost/credit_card.py
@@ -15,7 +15,7 @@ from easypost.resource import (
 
 class CreditCard(CreateResource, Resource):
     @classmethod
-    def fund(cls, api_key: Optional[str] = None, **params) -> "CreditCard":
+    def fund(cls, amount: str, primary_or_secondary: str = "primary", api_key: Optional[str] = None) -> "CreditCard":
         """Fund your EasyPost wallet by charging your primary or secondary card on file."""
         payment_methods = PaymentMethod.all()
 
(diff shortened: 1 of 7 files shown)


easypost/credit_card.py
```diff
@@ -0,0 +1,46 @@
...
+        primary_or_secondary = params.get("primary_or_secondary")
+        payment_method_to_use = payment_method_map.get(primary_or_secondary) if primary_or_secondary else None
+        card_id = payment_methods[payment_method_to_use]["id"] if payment_method_to_use else None
+
+        if payment_method_to_use is None or card_id is None or "card_" not in card_id:
+            raise Error(message="The chosen payment method is not a credit card. Please try again.")
+
+        requestor = Requestor(local_api_key=api_key)
+        url = f"{cls.class_url()}/{card_id}/charges"
+        wrapped_params = {"amount": params.get("amount")}
```
jchen293, line 36: I don't see the benefit of using `**params` in this function, can we just have two parameters in this function? 

Also In line 36, what if there is no `amount` in the `**params`? Should we check it before sending it to the API?

jchen293 (changes requested):

easypost/credit_card.py
```diff
@@ -0,0 +1,51 @@
...
+        card_id = payment_method_to_use.get("id")
+
+        requestor = Requestor(local_api_key=api_key)
+        url = f"{cls.class_url()}/{card_id}/charges"
+        wrapped_params = {"amount": params.get("amount")}
+        response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=wrapped_params)
+        return convert_to_easypost_object(response=response, api_key=api_key)
+
+    @classmethod
+    def delete(cls, easypost_id: str, api_key: Optional[str] = None):
```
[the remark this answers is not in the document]
Justintime50: I followed the pattern of the `retrieve` function which calls it `easypost_id`: https://github.com/EasyPost/easypost-python/blob/3f792eefd9468efb0ea2773aa72637398686b268/easypost/resource.py#L24, I'm not opposed to being more explicit here though

easypost/credit_card.py
Justintime50: Just following the pattern of the lib where we traditionally use `**params` for everything; however, I don't see the signature of this function changing much over time so we could use a more explicit keyword param approach.

Justintime50: As for checking if it's empty etc, I don't feel we should here as there are additional validations needed at the API level anyway. Let's pass it through and make the API catch what's necessary all in one go.

easypost/credit_card.py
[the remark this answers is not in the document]
jchen293: I prefer to be more explicit, similar to what we have done in the Referral class.

easypost/credit_card.py
jchen293: Got it. In Java, the param only needs `amount` so I wouldn't ask the user to input a Map in the parameter that only has one value.

easypost/credit_card.py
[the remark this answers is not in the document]
Justintime50: I've switched it

jchen293 (approved):

nwithan8 (approved):

PR #198, opened 2022-05-31
fix the credit card fund unit test parameters.

jchen293:
# Description

Fix the parameters in the credit card fund unit test

# Testing

n/a

# Pull Request Type

### DIFF
--- a/tests/test_credit_card.py
+++ b/tests/test_credit_card.py
@@ -6,10 +6,7 @@ import easypost
 @pytest.mark.vcr()
 @pytest.mark.skip("Skipping due to the lack of an available real credit card in tests")
 def test_credit_card_fund(prod_api_key):
-    credit_card = easypost.CreditCard.fund(
-        primary_or_secondary="primary",
-        amount="10",
-    )
+    credit_card = easypost.CreditCard.fund("100", "primary")
 
     assert credit_card
 


tests/test_credit_card.py
```diff
@@ -6,10 +6,7 @@
 @pytest.mark.vcr()
 @pytest.mark.skip("Skipping due to the lack of an available real credit card in tests")
 def test_credit_card_fund(prod_api_key):
-    credit_card = easypost.CreditCard.fund(
-        primary_or_secondary="primary",
-        amount="10",
-    )
+    credit_card = easypost.CreditCard.fund("100", "primary")
```
Justintime50, line 9: Heh, so Python has keyword arguments so leaving it as was should have technically worked.

Let's swap the order like you did but keep the explicit keywords in the params since Python supports it and will help tell users what is what. Let's also keep the amount low - in the off chance a user runs this (not us), we don't want to charge their card a large sum of money. (likelyhood is low as there's multiple other things they'd need to tweak to get there, but still we should protect where possible).

jchen293: sounds good!

Justintime50 (dismissed):

nwithan8 (approved):

PR #207, opened 2022-07-08
updated referral test

ryannguyen16:
# Description

changed the referral test all method to account for new referral_customers keys.

# Testing

updated cassettes. 

# Pull Request Type

### DIFF
--- a/tests/test_referral.py
+++ b/tests/test_referral.py
@@ -40,8 +40,11 @@ def test_referral_user_all(prod_api_key, page_size):
     """This test requires a partner user's production API key via EASYPOST_PROD_API_KEY."""
     referral_users = easypost.beta.Referral.all(page_size=page_size)
 
-    assert len(referral_users) <= page_size
-    assert all(isinstance(referral_user, easypost.User) for referral_user in referral_users)
+    referral_users_array = referral_users["referral_customers"]
+
+    assert len(referral_users_array) <= page_size
+    assert referral_users["has_more"] is not None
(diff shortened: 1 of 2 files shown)


Justintime50 (approved):
Nice!

jchen293 (approved):

tests/test_referral.py
```diff
@@ -40,8 +40,11 @@ def test_referral_user_all(prod_api_key, page_size):
     """This test requires a partner user's production API key via EASYPOST_PROD_API_KEY."""
     referral_users = easypost.beta.Referral.all(page_size=page_size)
 
-    assert len(referral_users) <= page_size
-    assert all(isinstance(referral_user, easypost.User) for referral_user in referral_users)
+    referral_users_array = referral_users["referral_customers"]
```
nwithan8, line 43: Why is this a JSON key rather than a property of an object?

nwithan8 (changes requested):

tests/test_referral.py
Justintime50: Because all of our tests are written this way?

nwithan8: I see, we don't have explicit "list" or "collection" wrapper objects, so we have to [look up a key in a dictionary](https://github.com/EasyPost/easypost-python/blob/5ede5f7a1ea4fa6ad9c20418eb190010ca89caf0/tests/test_shipment.py#L32). Ugh, this library.

nwithan8 (approved):

PR #213, opened 2022-07-20
feat: adds validate webhook function

Justintime50:
# Description

Adds a new function to validate if a webhook originated from EasyPost with the stored Webhook secret, if they do the function will return True, otherwise False.

# Testing

- Adds a unit test to ensure this works by checking that the dummy webhook and secret match the expected hash for that same data.

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## NEXT RELEASE
+
+- Adds `validate_webhook` function that returns your webhook or raises an error if there is a `webhook_secret` mismatch
+
 ## v7.3.0 (2022-07-18)
 
 - Adds ability to generate forms for shipments via `generate_form()` function
--- a/easypost/webhook.py
+++ b/easypost/webhook.py
@@ -1,3 +1,13 @@
+import hashlib
+import hmac
+import json
+import unicodedata
+from typing import (
+    Any,
+    Dict,
+)
+
+from easypost.error import Error
 from easypost.requestor import (
     RequestMethod,
     Requestor,
@@ -17,3 +27,35 @@ class Webhook(AllResource, CreateResource, DeleteResource):
         response, api_key = requestor.request(method=RequestMethod.PATCH, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    @staticmethod
+    def validate_webhook(event_body: bytes, headers: Dict[str, Any], webhook_secret: str) -> Dict[str, Any]:
+        """Validate a webhook by comparing the HMAC signature header sent from EasyPost to your shared secret.
+        If the signatures do not match, an error will be raised signifying the webhook either did not originate
+        from EasyPost or the secrets do not match. If the signatures do match, the `event_body` will be returned
+        as JSON.
+        """
+        easypost_hmac_signature = headers.get("X-Hmac-Signature")
+
+        if easypost_hmac_signature:
+            normalized_secret = unicodedata.normalize("NFKD", webhook_secret)
+            encoded_secret = bytes(normalized_secret, "utf8")
+
+            expected_signature = hmac.new(
+                key=encoded_secret,
+                msg=event_body,
+                digestmod=hashlib.sha256,
+            )
+
+            digest = "hmac-sha256-hex=" + expected_signature.hexdigest()
+
+            if hmac.compare_digest(digest, easypost_hmac_signature):
+                webhook_body = json.loads(event_body)
+            else:
+                raise Error(
+                    message="Webhook received did not originate from EasyPost or had a webhook secret mismatch."
+                )
+        else:
+            raise Error(message="Webhook received does not contain an HMAC signature.")
+
+        return webhook_body
--- a/tests/test_webhook.py
+++ b/tests/test_webhook.py
@@ -54,3 +54,51 @@ def test_webhook_delete(webhook_url):
 
     # This endpoint/method does not return anything, just make sure the request doesn't fail
     assert isinstance(response, easypost.Webhook)
+
+
+def test_validate_webhook(event_body):
+    webhook_secret = "sécret"
+    expected_hmac_signature = "hmac-sha256-hex=e93977c8ccb20363d51a62b3fe1fc402b7829be1152da9e88cf9e8d07115a46b"
+    headers = {
+        "X-Hmac-Signature": expected_hmac_signature,
+    }
+
+    webhook_body = easypost.Webhook.validate_webhook(
+        event_body=event_body,
+        headers=headers,
+        webhook_secret=webhook_secret,
+    )
+
+    assert webhook_body["description"] == "batch.created"
+
+
+def test_validate_webhook_invalid_secret(event_body):
+    webhook_secret = "invalid_secret"
+    headers = {
+        "X-Hmac-Signature": "some-signature",
+    }
+
+    with pytest.raises(easypost.Error) as error:
+        _ = easypost.Webhook.validate_webhook(
+            event_body=event_body,
+            headers=headers,
+            webhook_secret=webhook_secret,
+        )
+
+    assert str(error.value) == "Webhook received did not originate from EasyPost or had a webhook secret mismatch."
+
+
+def test_validate_webhook_missing_secret(event_body):
+    webhook_secret = "123"
+    headers = {
+        "some-header": "some-value",
+    }
+
+    with pytest.raises(easypost.Error) as error:
+        _ = easypost.Webhook.validate_webhook(
+            event_body=event_body,
+            headers=headers,
+            webhook_secret=webhook_secret,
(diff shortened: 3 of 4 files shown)


easypost/webhook.py
```diff
@@ -17,3 +27,28 @@ def update(self, **params) -> "Webhook":
...
+    @staticmethod
+    def validate_webhook(webhook_body: Dict[str, Any], headers: Dict[str, Any], webhook_secret: str) -> bool:
+        """Validate a webhook originated from EasyPost by comparing the HMAC header to a shared secret.
+        This function will return True if the webhook originated from EasyPost and False if it did not.
+        """
+        secret_validated = False
+        easypost_signature = headers.get("x-hmac-signature")
+
+        if easypost_signature:
+            secret = binascii.hexlify(bytes(unicodedata.normalize("NFKD", webhook_secret), "utf8"))
```
nwithan8, line 40: Please break this line up to make it easier to read.

nwithan8 (changes requested):

easypost/webhook.py
Justintime50: Broken up

jchen293 (approved):

nwithan8 (approved):

easypost/webhook.py
```diff
@@ -17,3 +27,30 @@ def update(self, **params) -> "Webhook":
         response, api_key = requestor.request(method=RequestMethod.PATCH, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    @staticmethod
+    def validate_webhook(webhook_body: Dict[str, Any], headers: Dict[str, Any], webhook_secret: str) -> bool:
```
Roguelazer, line 32: you can't validate over a dict, you can only validate over a series of bytes, because JSON doesn't have a canonical serialization.

easypost/webhook.py
```diff
@@ -17,3 +27,30 @@ def update(self, **params) -> "Webhook":
...
+            expected_signature = hmac.new(
+                key=hexified_secret,
+                msg=webhook_bytes,
+                digestmod=hashlib.sha256,
+            )
+
+            digest = "hmac-sha256-hex=" + expected_signature.hexdigest()
+
+            if hmac.compare_digest(digest, easypost_signature):
+                secret_validated = True
```
Roguelazer, line 54: generally my preference for an API design is to raise an exception on mismatch, instead of returning a boolean. That makes it harder to do the insecure thing.

Roguelazer (changes requested):

easypost/webhook.py
Justintime50: Ah yes, had this in the function when really this should have happened in the fixture and passed into the function. Corrected.

easypost/webhook.py
Justintime50: Doh, that would be the natural thing wouldn't it? Thanks for the callout. It now raises an exception instead.

easypost/webhook.py
```diff
@@ -29,28 +30,33 @@ def update(self, **params) -> "Webhook":
         return self
 
     @staticmethod
-    def validate_webhook(webhook_body: Dict[str, Any], headers: Dict[str, Any], webhook_secret: str) -> bool:
+    def validate_webhook(event_body: bytes, headers: Dict[str, Any], webhook_secret: str) -> Dict[str, Any]:
         """Validate a webhook originated from EasyPost by comparing the HMAC header to a shared secret.
         This function will return True if the webhook originated from EasyPost and False if it did not.
         """
-        secret_validated = False
+        webhook_body = {}
```
Roguelazer, line 37: this is dead code, right?

Justintime50: Indeed, replaced by the else statement. Removed.

easypost/webhook.py
```diff
@@ -17,3 +28,35 @@ def update(self, **params) -> "Webhook":
...
+                msg=event_body,
+                digestmod=hashlib.sha256,
+            )
+
+            digest = "hmac-sha256-hex=" + expected_signature.hexdigest()
+
+            if hmac.compare_digest(digest, easypost_signature):
+                webhook_body = json.loads(event_body)
+            else:
+                raise Error(
```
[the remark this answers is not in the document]
Justintime50: I don't disagree; however, we only ever use this EasyPost Error class throughout the library at the moment. We have plans to rework error handling in an upcoming major release; but current users have come to expect this pattern for now.

easypost/webhook.py
```diff
@@ -17,3 +28,34 @@ def update(self, **params) -> "Webhook":
         response, api_key = requestor.request(method=RequestMethod.PATCH, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
+
+    @staticmethod
+    def validate_webhook(event_body: bytes, headers: Dict[str, Any], webhook_secret: str) -> Dict[str, Any]:
+        """Validate a webhook originated from EasyPost by comparing the HMAC header to a shared secret.
+        This function will return True if the webhook originated from EasyPost and False if it did not.
```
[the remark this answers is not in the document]
Justintime50: Great catch! The docstring is old as we tweaked the logic a bit. I've corrected the docstring to align better with what's actually happening.

easypost/webhook.py
```diff
@@ -17,3 +28,34 @@ def update(self, **params) -> "Webhook":
...
+        """
+        easypost_signature = headers.get("x-hmac-signature")
+
+        if easypost_signature:
```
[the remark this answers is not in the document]
Justintime50: Great points. I've simplified the function to ensure the existence of the header prior to returning the event body and raising an error if there is no header. I feel if someone wanted a non-strict mode they'd just parse the JSON themselves outside this function so we can keep this explicit. I added another unit test to verify that branch of logic fires correctly.

Justintime50:
@Roguelazer @hagemt this is ready for re-review when you're able.

Roguelazer (approved):

easypost/webhook.py
```diff
@@ -17,3 +28,36 @@ def update(self, **params) -> "Webhook":
...
+            if hmac.compare_digest(digest, easypost_hmac_signature):
+                webhook_body = json.loads(event_body)
+            else:
+                raise Error(
+                    message="Webhook received did not originate from EasyPost or had a webhook secret mismatch."
+                )
+        else:
+            raise Error(message="Webhook received does not contain an HMAC signature.")
+
+        return webhook_body
```
nwithan8, line 63: Still not a fan that this function does two separate things, but I won't block it.

nwithan8 (approved):

easypost/webhook.py
Justintime50: It validates your webhook then gives it to you if there is no error, it's not doing two things. You send your webhook in and get it out barring no error.

nwithan8: > It validates your webhook then gives it to you if there is no error

Exactly, it's doing one thing (validating it) and then doing another thing (deserializing the JSON; what if the webhook is ever not JSON?)

Deserializing the webhook should be done in a separate, explicit step by the end-user.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## NEXT RELEASE
+
+- Adds `validate_webhook` function that returns True/False depending on if a webhook originated from EasyPost
```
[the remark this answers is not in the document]
Justintime50: I'll fix this prior to us releasing. Don't want to remove all the approvals. Good callout

easypost/webhook.py
Justintime50: This decision was made from feedback on an internal discussion:

> what james is saying that the body goes in as a series of bytes, and what comes out is a deserialized structure; you want to build the signature validation into the routine that deserializes. If the signature validation fails, it raises an exception. This way you don't end up with a state where you have a deserialized structure that is the result of an invalid signature

Justintime50:
Need to tweak the test fixture as this isn't actually working as intended. Drafting.

tests/test_webhook.py
```diff
@@ -54,3 +54,51 @@ def test_webhook_delete(webhook_url):
...
+        _ = easypost.Webhook.validate_webhook(
+            event_body=event_body,
+            headers=headers,
+            webhook_secret=webhook_secret,
+        )
+
+    assert str(error.value) == "Webhook received did not originate from EasyPost or had a webhook secret mismatch."
+
+
+def test_validate_webhook_missing_secret(event_body):
```
jchen293, line 91: This test function name needs to be renamed, it's testing missing `HMAC signature header`

Justintime50: It's named properly. Look at the `headers` variable, there is no `X-Hmac-Signature` entry from EasyPost.

jchen293 (approved):

nwithan8 (approved):

PR #214, opened 2022-07-21
add ability to create/buy a shipment that has carbon_offset

jchen293:
# Description

- Adds the ability to create a shipment with carbon_offset
- Adds the ability to buy a shipment with carbon_offset
- Adds the ability to one-call-buy a shipment with carbon_offset
- Adds the ability to rerate a shipment with carbon_offset

# Testing

Added three unit tests and passed it

# Pull Request Type

### DIFF
--- a/easypost/shipment.py
+++ b/easypost/shipment.py
@@ -25,8 +25,7 @@ class Shipment(AllResource, CreateResource):
         """Create an Shipment object."""
         requestor = Requestor(local_api_key=api_key)
         url = cls.class_url()
-        wrapped_params = {cls.snakecase_name(): params}
-        wrapped_params["carbon_offset"] = carbon_offset  # type: ignore
+        wrapped_params = {cls.snakecase_name(): params, "carbon_offset": carbon_offset}
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=wrapped_params)
         return convert_to_easypost_object(response=response, api_key=api_key)
 
--- a/easypost/shipment.py
+++ b/easypost/shipment.py
@@ -6,6 +6,7 @@ from typing import (
 )
 
 from easypost import Rate
+from easypost.easypost_object import convert_to_easypost_object
 from easypost.error import Error
 from easypost.requestor import (
     RequestMethod,
@@ -19,11 +20,27 @@ from easypost.util import get_lowest_object_rate
 
 
 class Shipment(AllResource, CreateResource):
-    def regenerate_rates(self) -> "Shipment":
+    @classmethod
+    def create(cls, api_key: Optional[str] = None, carbon_offset: bool = False, **params) -> Any:
+        """Create an Shipment object."""
+        requestor = Requestor(local_api_key=api_key)
+        url = cls.class_url()
+        wrapped_params = {cls.snakecase_name(): params}
+        if carbon_offset:
+            wrapped_params["carbon_offset"] = carbon_offset  # type: ignore
+        response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=wrapped_params)
+        return convert_to_easypost_object(response=response, api_key=api_key)
+
+    def regenerate_rates(self, carbon_offset=False) -> "Shipment":
         """Regenerate rates for a shipment."""
         requestor = Requestor(local_api_key=self._api_key)
         url = "%s/%s" % (self.instance_url(), "rerate")
-        response, api_key = requestor.request(method=RequestMethod.POST, url=url)
+        params = {}
+        if carbon_offset:
+            params = {
+                "carbon_offset": carbon_offset,
+            }
+        response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
 
@@ -34,10 +51,13 @@ class Shipment(AllResource, CreateResource):
         response, _ = requestor.request(method=RequestMethod.GET, url=url)
         return response.get("result", [])
 
-    def buy(self, **params) -> "Shipment":
+    def buy(self, carbon_offset=False, **params) -> "Shipment":
         """Buy a shipment."""
         requestor = Requestor(local_api_key=self._api_key)
         url = "%s/%s" % (self.instance_url(), "buy")
+        if carbon_offset:
+            params["carbon_offset"] = carbon_offset
+
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=params)
         self.refresh_from(values=response, api_key=api_key)
         return self
--- a/easypost/shipment.py
+++ b/easypost/shipment.py
@@ -26,7 +26,7 @@ class Shipment(AllResource, CreateResource):
         requestor = Requestor(local_api_key=api_key)
         url = cls.class_url()
         wrapped_params = {cls.snakecase_name(): params}
-        wrapped_params["carbon_offset"] = carbon_offset
+        wrapped_params["carbon_offset"] = carbon_offset  # type: ignore
         response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=wrapped_params)
         return convert_to_easypost_object(response=response, api_key=api_key)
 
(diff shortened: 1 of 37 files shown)


easypost/shipment.py
```diff
@@ -19,11 +20,27 @@
 
 
 class Shipment(AllResource, CreateResource):
-    def regenerate_rates(self) -> "Shipment":
+    @classmethod
+    def create(cls, api_key: Optional[str] = None, carbon_offset: bool = False, **params) -> Any:
```
Justintime50, line 24: This does not return anything, it returns a Shipment object.

```suggestion
    def create(cls, api_key: Optional[str] = None, carbon_offset: bool = False, **params) -> "Shipment":
```

easypost/shipment.py
```diff
@@ -19,11 +20,27 @@
...
+        """Create an Shipment object."""
+        requestor = Requestor(local_api_key=api_key)
+        url = cls.class_url()
+        wrapped_params = {cls.snakecase_name(): params}
+        if carbon_offset:
```
Justintime50, line 29: Let's remove the `if` statement and always pass in carbon_offset. If the user did not request it, it will be set to False by default.

easypost/shipment.py
```diff
@@ -19,11 +20,27 @@
...
+            wrapped_params["carbon_offset"] = carbon_offset  # type: ignore
+        response, api_key = requestor.request(method=RequestMethod.POST, url=url, params=wrapped_params)
+        return convert_to_easypost_object(response=response, api_key=api_key)
+
+    def regenerate_rates(self, carbon_offset=False) -> "Shipment":
```
Justintime50, line 34:
```suggestion
    def regenerate_rates(self, carbon_offset: bool = False) -> "Shipment":
```

easypost/shipment.py
```diff
@@ -19,11 +20,27 @@
...
         """Regenerate rates for a shipment."""
         requestor = Requestor(local_api_key=self._api_key)
         url = "%s/%s" % (self.instance_url(), "rerate")
-        response, api_key = requestor.request(method=RequestMethod.POST, url=url)
+        params = {}
+        if carbon_offset:
```
Justintime50, line 39: Let's always pass carbon offset and remove the if statement. If the user doesn't want it, it will get set to False.

easypost/shipment.py
```diff
@@ -34,10 +51,13 @@ def get_smartrates(self) -> List[object]:
         response, _ = requestor.request(method=RequestMethod.GET, url=url)
         return response.get("result", [])
 
-    def buy(self, **params) -> "Shipment":
+    def buy(self, carbon_offset=False, **params) -> "Shipment":
```
Justintime50, line 54:
```suggestion
    def buy(self, carbon_offset: bool = False, **params) -> "Shipment":
```

easypost/shipment.py
```diff
@@ -34,10 +51,13 @@ def get_smartrates(self) -> List[object]:
...
         """Buy a shipment."""
         requestor = Requestor(local_api_key=self._api_key)
         url = "%s/%s" % (self.instance_url(), "buy")
+        if carbon_offset:
```
Justintime50, line 58: Let's always pass carbon_offset.

tests/test_shipment.py
```diff
@@ -244,3 +244,34 @@ def test_generate_form(one_call_buy_shipment, rma_form_options):
...
+    assert all(rate.carbon_offset is not None for rate in shipment.rates)
+
+
+@pytest.mark.vcr()
+def test_shipment_rerate_with_carbon_offset(carbon_offset_shipment_one_call_buy):
+    shipment = easypost.Shipment.create(**carbon_offset_shipment_one_call_buy)
+
+    rates = shipment.regenerate_rates(carbon_offset=True)
+
+    assert all(rate.carbon_offset is not None for rate in rates.rates)
```
Justintime50, line 277: Please mirror the tests from the Java implementation of Carbon Offset as they have much stronger assertions than this.

Justintime50 (changes requested):

easypost/shipment.py
jchen293: Yes, we could. This will impact about 25 unit tests and we will need to re-record the cassettes

Justintime50: Let's do it. I think it's what we decided for the other libs too.

tests/test_shipment.py
jchen293: Added a few more, using `all` in python assert is much shorter than Java

tests/test_shipment.py
```diff
@@ -211,10 +211,10 @@ def test_shipment_get_lowest_smartrate(basic_shipment):
 
     # Test lowest smartrate with valid filters
     lowest_smartrate_filters = easypost.Shipment.get_lowest_smartrate(
-        smartrates, delivery_days=1, delivery_accuracy="percentile_90"
+        smartrates, delivery_days=2, delivery_accuracy="percentile_90"
```
jchen293, line 214: ## Reviewer Note:

After re-recording the cassettes, the smart rate delivery day got changed, update of the unit test is needed.

easypost/shipment.py
jchen293: removed

tests/test_shipment.py
```diff
@@ -244,3 +244,42 @@ def test_generate_form(one_call_buy_shipment, rma_form_options):
...
+
+    assert isinstance(shipment, easypost.Shipment)
+    assert shipment.rates is not None
+    assert all(rate.carbon_offset is not None for rate in shipment.rates)
+
+
+@pytest.mark.vcr()
+def test_shipment_buy_with_carbon_offset(carbon_offset_shipment):
+    shipment = easypost.Shipment.create(**carbon_offset_shipment)
+    shipment.buy(carbon_offset=True, rate=shipment.lowest_rate())
```
Justintime50, line 261: Please swap these orders so that they come in order of importance. Rate first, carbon_offset second

tests/test_shipment.py
```diff
@@ -244,3 +244,42 @@ def test_generate_form(one_call_buy_shipment, rma_form_options):
...
+@pytest.mark.vcr()
+def test_shipment_rerate_with_carbon_offset(carbon_offset_shipment_one_call_buy):
+    shipment = easypost.Shipment.create(**carbon_offset_shipment_one_call_buy)
+
+    new_carbon_rates = shipment.regenerate_rates(carbon_offset=True)
+
+    assert all(rate.carbon_offset is not None for rate in new_carbon_rates.rates)
+
+    new_non_carbon_rates = shipment.regenerate_rates()
+    with pytest.raises(AttributeError):
```
Justintime50, line 284: Why are we raising an AttributeError here?

jchen293: Because we are regenerating rate without `carbon_offset` and there will be no `carbon_offset` attribute in the `rate` object which will throw `AttributeError`, I don't think this test is necessary here because we have a specific unit test for `regenerate_rates()`, thought?

tests/test_shipment.py
jchen293: good catch

tests/test_shipment.py
nwithan8: I do a similar thing in the Ruby unit test, since, as Jack said, the `carbon_offset` property simply won't exist, so we need to catch the error of trying to access a non-existent property.

Please advise, whatever we do here we should do for Ruby (and I guess Node and PHP?)

easypost/shipment.py
```diff
@@ -19,11 +20,24 @@
...
 
 class Shipment(AllResource, CreateResource):
-    def regenerate_rates(self) -> "Shipment":
+    @classmethod
+    def create(cls, api_key: Optional[str] = None, carbon_offset: Optional[bool] = False, **params) -> "Shipment":
+        """Create an Shipment object."""
+        requestor = Requestor(local_api_key=api_key)
+        url = cls.class_url()
+        wrapped_params = {cls.snakecase_name(): params}
+        wrapped_params["carbon_offset"] = carbon_offset  # type: ignore
```
Justintime50, line 29: This should fix the need to ignore the type while providing type safety.

```suggestion
        wrapped_params = {cls.snakecase_name(): params}  # type: Dict[str, Any]
        wrapped_params["carbon_offset"] = carbon_offset
```

Justintime50 (approved):

nwithan8 (approved):

PR #218, opened 2022-07-29
fix: looser verify and verify_strict param type allowance

Justintime50:
# Description

Allows for looser `verify` and `verify_strict` params because strings and bools are valid syntax (which is what we have documented). Also corrects the type hints for these. 

# Testing

- Updates old tests to use booleans to match our docs
- Adds a new test to ensure array syntax still works. 

# Pull Request Type

### DIFF
--- a/easypost/address.py
+++ b/easypost/address.py
@@ -2,6 +2,7 @@ from typing import (
     Any,
     Dict,
     Optional,
+    Union,
 )
 
 from easypost.easypost_object import convert_to_easypost_object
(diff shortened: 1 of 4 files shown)


easypost/address.py
```diff
@@ -20,21 +21,19 @@ class Address(AllResource, CreateResource):
     def create(
         cls,
         api_key: Optional[str] = None,
-        verify: Optional[Dict[str, Any]] = None,
-        verify_strict: Dict[str, Any] = None,
+        verify: Optional[Union[Dict[str, Any], str, bool]] = None,
```
nwithan8, line 24: Thought: I hate this. Shouldn't we be driving people to only providing a boolean (especially in Python, where this type hint is just a suggestion and not actually enforced)?

nwithan8 (changes requested):

easypost/address.py
Justintime50: We can't break backwards compatibility or expectations. But yes, over a long period of time we can. First step is updating the docs to say "use a bool" which is done. Then in an upcoming major release we could start restricting this. But almost everyone is using arrays.

nwithan8: I understand. I was just saying, we could start "restricting" it now, since it's not actually restricted (the type hint can say "you have to provide a boolean", but users could continue giving it an array and it won't actually break, yay non-type-strict languages)

nwithan8 (approved):

easypost/address.py
Justintime50: Ah. Well because we provide these type hints to users who import the library, it would complain if they also use mypy which would then force them to ignore a warning (not ideal) or make a code change on a non-major version bump.

And also, even though we may only "want" users to do a bool, the others are still valid options.

PR #233, opened 2022-09-30
feat: adds cassette expirations

Justintime50:
# Description

Adds a test utility that checks for expired cassettes on each run and throws errors if cassettes must be re-recorded due to old age to ensure the data is always up-to-date within a half year.

# Testing

I set the days to ~40 and errors were thrown, setting it to 180 has the suite pass 

# Pull Request Type

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import os
 from typing import (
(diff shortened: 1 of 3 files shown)


Justintime50:
Blocked until #232 is merged

tests/conftest.py
```diff
@@ -69,9 +70,26 @@ def partner_prod_api_key():
...
+        if cassette_age < expiration_age:
+            raise Exception(
+                f"{cassette_filepath} is older than {expiration_days} and has expired. Please re-record the cassette."
+            )
+
+
 @pytest.fixture(scope="session")
 def vcr_config():
     """Setup the VCR config for the test suite."""
+    check_expired_cassettes()
```
nwithan8, line 91: Nit: I prefer to be explicit on parameters, even if it's the same as the default value. Just makes it more readable.

tests/conftest.py
```diff
@@ -69,9 +70,26 @@ def partner_prod_api_key():
...
 
 
+def check_expired_cassettes(expiration_days: int = 180):
+    """Checks for expired cassettes and throws errors if they are too old and must be re-recorded."""
+    test_name = os.environ.get("PYTEST_CURRENT_TEST").split(":")[-1].split(" ")[0]  # type: ignore
+    cassette_filepath = os.path.join("tests", "cassettes", f"{test_name}.yaml")
+
+    if os.path.exists(cassette_filepath):
+        expiration_age = datetime.datetime.now() - datetime.timedelta(days=expiration_days)
+        cassette_age = datetime.datetime.fromtimestamp(os.stat(cassette_filepath).st_mtime)
```
nwithan8, line 80: Question: Is there no way to store a timestamp inside the file rather than relying on the file metadata? Moving the file would change the timestamp, even though the contents aren't changed.

nwithan8 (dismissed):

tests/conftest.py
Justintime50: This is no longer applicable

tests/conftest.py
Justintime50: I checked the docs for `vcrpy` and there's no mention of a timestamp sadly. They don't include one we could use either.

jontsai (approved):

nwithan8 (approved):

PR #242, opened 2022-11-02
feat: adds library documentation

Justintime50:
# Description

Adds library documentation and build targets

# Testing

# Pull Request Type


jchen293 (dismissed):

jchen293 (approved):

nwithan8 (approved):
I prefer Sphinx, but pdoc seems to do the trick

Justintime50:
> I prefer Sphinx, but pdoc seems to do the trick

Sphinx is definitely the better standard, though it takes way too much config. Really we just need something that can run out of the box with a single command. If we decide in the future we want more out of these docs we can swap to Sphinx for sure.

PR #245, opened 2022-11-22
Handle adding carrier accounts that require custom workflows

nwithan8:
# Description

- Overload existing `create` function in carrier_account to dynamically adjust URL based on carrier account type

# Testing

- Update unit test, cassette

# Pull Request Type


easypost/carrier_account.py
```diff
@@ -16,6 +17,19 @@
 )
 
 
+def _select_carrier_account_creation_endpoint(carrier_account_type: Optional[Any]) -> str:
+    """Determines which API endpoint to use for the creation call."""
+    carriers_with_custom_workflows = [
+        "FedExAccount",
+        "UpsAccount",
+    ]
```
Justintime50, line 25: These should probably be constants in each lib where possible? Small performance improvements.

easypost/carrier_account.py
```diff
@@ -24,11 +38,16 @@ def types(cls, api_key: Optional[str] = None) -> List[str]:
...
 
+    # Overrides CreateResource.create
     @classmethod
-    def register(cls, api_key: Optional[str] = None, **params) -> "CarrierAccount":
+    def create(cls, api_key: Optional[str] = None, **params) -> "CarrierAccount":
         """Creates a Carrier Account that requires custom registration (eg: FedEx & UPS)."""
         requestor = Requestor(local_api_key=api_key)
-        url = f"{cls.class_url()}/register"
+
+        if params.get("type") is None:
```
Justintime50, line 47: mAkE iT pYtHoNiC 😉 

`params.get("type")` is defined twice in this function, let's assign it to a variable and reuse that.

nwithan8: Resolved

easypost/carrier_account.py
nwithan8: Resolved

easypost/constant.py
```diff
@@ -1,2 +1,5 @@
 SUPPORT_EMAIL = "support@easypost.com"
 TIMEOUT = 60
+
+FEDEX_ACCOUNT_TYPE = "FedExAccount"
+UPS_ACCOUNT_TYPE = "UpsAccount"
```
Justintime50, line 5: Heh, I meant this should be a constant of type list containing the strings individually. Much easier than building multiple constants down the stack. One place of maintenance (inside the list constant) that we can then simply check against is much better than multiple single constants that we later have to build manually.

nwithan8: Counter: If it's constants like this, the end-user can use them when forming their data (for example, "is it supposed to be Ups or UPS")?

Justintime50: Although that would be nice, that starts to get into the weeds of "how much maintenance cost do we want to deal with" which the answer is still "as little as possible". We shouldn't expose something unless we are prepared to maintain it. If the constant name is clear we are using it for internal validation for this function, it will dissuade users from using it.

easypost/constant.py
```diff
@@ -1,2 +1,7 @@
 SUPPORT_EMAIL = "support@easypost.com"
 TIMEOUT = 60
+
+FEDEX_ACCOUNT_TYPE = "FedexAccount"
+UPS_ACCOUNT_TYPE = "UpsAccount"
+
+_CUSTOM_WORKFLOW_CARRIER_ACCOUNT_TYPES = [FEDEX_ACCOUNT_TYPE, UPS_ACCOUNT_TYPE]
```
jchen293, line 7: NIT: no need to define the variables since they are only being used once, I would pass them in the list directly

easypost/constant.py
nwithan8: We'd have to maintain it if the API ever changes anyway, this at least kills two birds with one stone?

easypost/constant.py
nwithan8: See: https://github.com/EasyPost/easypost-python/pull/245#discussion_r1029912613

jchen293 (approved):

easypost/constant.py
nwithan8: Per discussion elsewhere, this will be tweaked

jchen293 (dismissed):

easypost/carrier_account.py
```diff
@@ -16,6 +18,14 @@
 )
 
 
+def _select_carrier_account_creation_endpoint(carrier_account_type: Optional[Any]) -> str:
+    """Determines which API endpoint to use for the creation call."""
+    if carrier_account_type in _CARRIER_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS:
```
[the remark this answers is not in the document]
nwithan8: Done to suggest it's internal (like a private method)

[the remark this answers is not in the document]
Justintime50: Right, I understand why but meh for constants.

They no longer match. My suggestion didn't have the underscore in it so you'll need to adjust one or the other to match since this var is now undefined.

Justintime50 (approved):

jchen293 (approved):

PR #258, opened 2023-02-13
fix: use build package for build process

Justintime50:
# Description

Installing/building via `setup.py` directly is deprecated. We are migrating to the `build` package to do the same thing for us. I inspected the output package and everything is the same as before (`sdist` and `wheels` are built as before with a simpler interface and include all the same items as before. I've had great success with this package on dozens of personal projects over the last year.

# Testing

Ran `make build` and got the output of `Successfully built easypost-7.9.0.tar.gz and easypost-7.9.0-py3-none-any.whl`, the added benefit of `build` is it creates a temporary virtual env just for this build which should isolate this flow even better.

# Pull Request Type


.gitignore
```diff
@@ -1,32 +1,25 @@
-.env
+._*
```
Justintime50, line 1: # Reviewer Note

Had to tweak the gitignore to change for the build output, sorted the list and cleaned it up while here.

pyproject.toml
```diff
@@ -3,3 +3,11 @@ line-length = 120
 
 [tool.isort]
 profile = "black"
+line_length = 120
```
Justintime50, line 6: # Reviewer Note

Consolidating this from its previously separate file into this one

jchen293 (approved):

nwithan8 (approved):

PR #266, opened 2023-04-26
feat: adds retrieve_estimated_delivery_date function

Justintime50:
# Description

- Adds `retrieve_estimated_delivery_date` function to the Shipment class

# Testing

Adds a unit test to ensure the correct keys are in the response

# Pull Request Type

### DIFF
--- a/easypost/shipment.py
+++ b/easypost/shipment.py
@@ -161,3 +161,13 @@ class Shipment(NextPageResource):
             raise Error(message="No rates found.")
 
         return lowest_smartrate
+
+    def retrieve_estimated_delivery_date(self, planned_ship_date: str) -> List[Dict[str, Any]]:
+        """Retrieves the estimated delivery date of each Rate via SmartRate."""
+        requestor = Requestor(local_api_key=self._api_key)
+        url = "%s/%s" % (self.instance_url(), "smartrate/delivery_date")
+        params = {
+            "planned_ship_date": planned_ship_date,
+        }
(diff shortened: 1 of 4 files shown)


nwithan8 (approved):

easypost/shipment.py
```diff
@@ -161,3 +161,13 @@ def get_lowest_smartrate(smartrates, delivery_days: int, delivery_accuracy: str)
             raise Error(message="No rates found.")
 
         return lowest_smartrate
+
+    def retrieve_estimated_delivery_date(self, planned_ship_date: str) -> List[Dict[str, Any]]:
```
nwithan8, line 165: Function returns rate objects, correct? Function name misleading.

jchen293 (approved):

easypost/shipment.py
Justintime50: Kind of, it returns the original rate objects (nested) with a new `easypost_time_in_transit_data` object attached to each so it's not an actual list of Rate objects because it's a list of objects that contain both a `rate` and an `easypost_time_in_transit_data` object inside them.

As for the function name, it matches the API docs provided which is the get the estimated delivery date for rates. Does that work? https://www.easypost.com/docs/api#retrieve-estimated-delivery-date-and-total-transit-days-across-all-rates-for-a-shipment

nwithan8: Sounds good

PR #273, opened 2023-05-11
easypost client

Justintime50:
# Description

Adds the new `EasyPostClient` object allowing for thread-safety. The repo is in a hybrid stage where the old way of doing things still exists but we're able to introduce the new approach as well. In follow-up PRs, I'll add the remaining services/models, eventually remove the old code for each, then cleanup the requester and all related code for actually making requests. This allows the current test suite to remain for now while letting us switch over to the new approach.

# Testing

- [x] Added new tests for the Client object
- [x] Updated syntax for address tests 

# Pull Request Type


easypost/requestor.py
```diff
@@ -227,7 +230,7 @@ def requests_request(
                 params=url_params,
                 headers=headers,
                 json=body,
-                timeout=timeout,
+                timeout=self.client.timeout if self.client else timeout,  # TODO: Remove fallback
```
jchen293, line 233: question: I know we previously only define one timeout, can we have a tuple to set up the connection and open timeouts? Below is the link I was reading.

https://requests.kennethreitz.org/en/latest/user/advanced/#timeouts

easypost/services/address_service.py
```diff
@@ -0,0 +1,73 @@
...
+    def create(
+        self,
+        verify: Optional[Union[Dict[str, Any], str, bool]] = None,
+        verify_strict: Optional[Union[Dict[str, Any], str, bool]] = None,
+        **params,
+    ) -> Address:
+        """Create an Address."""
+        url = self.class_url(self.model_class)
+
+        wrapped_params = {self.snakecase_name(self.model_class): params}  # type: Dict[str, Any]
```
jchen293, line 31: Suggestion: `self.snakecase_name` here is redundant because `model_class` is only a single word

easypost/services/base_service.py
```diff
@@ -0,0 +1,101 @@
...
+            return f"/{class_name}s"
+
+    def instance_url(self, class_name, id) -> str:
+        """Generate an instance URL based on the ID of the object."""
+        if not id:
+            raise Error(f"{class_name} instance has invalid ID: {id}")
+
+        class_url = self.class_url(class_name)
+
+        return f"{class_url}/{id}"
```
jchen293, line 41: Issue: missing type hints on these functions

easypost/services/base_service.py
```diff
@@ -0,0 +1,101 @@
...
+        response, api_key = Requestor().request(method=RequestMethod.GET, url=url, params=params)
+
+        # TODO: Get rid of the api_key
+        return convert_to_easypost_object(response=response, api_key=api_key)
+
+    def retrieve_resource(self, class_name: str, id: str) -> Any:
+        """Retrieve an object from the EasyPost API."""
+        url = self.instance_url(class_name, id)
+        # TODO: Don't instantiate the Requestor class
+        response, api_key = Requestor().request(method=RequestMethod.GET, url=url)
```
jchen293, line 67: Issue: There is a trade-off in making these general functions (create, retrieve, all) in the parent class, this will allow users to use these functions on a service that doesn't support it.

easypost/services/address_service.py
```diff
@@ -0,0 +1,73 @@
...
+        if verify:
+            wrapped_params["verify"] = verify
+        if verify_strict:
+            wrapped_params["verify_strict"] = verify_strict
+
+        response, api_key = Requestor().request(method=RequestMethod.POST, url=url, params=wrapped_params)
+
+        return convert_to_easypost_object(response=response, api_key=api_key)
```
jchen293, line 39: Question: Can we remove the `api_key` as part of this PR?

jchen293 (changes requested):

easypost/services/base_service.py
Justintime50: Will fix all type hints in a follow up PR

easypost/services/address_service.py
Justintime50: No, all the old functions still need it. I'll be cleaning that all up once I can delete the old logic

tests/test_easypost_client.py
```diff
@@ -0,0 +1,35 @@
...
+    assert client2.api_key == "456"
+
+
+def test_api_base():
+    """Tests that we can override the API base of the client object."""
+    client1 = EasyPostClient(api_key="123")
+    assert client1.api_base == "https://api.easypost.com/v2"
+
+    client2 = EasyPostClient(api_key="123", api_base="http://example.com")
+    assert client2.api_base == "http://example.com"
```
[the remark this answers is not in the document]
Justintime50: api_base includes the version and ultimately it doesn't matter, users won't really use the `api_base` variable, it's used internally by engineers if they wanted to point to their local dev API

easypost/services/base_service.py
Justintime50: I can make them "internal" by prepending an underscore? There isn't really a great way in Python to make something private and not accessible to the users.

easypost/services/address_service.py
Justintime50: I'm going to apply the snakecase to everything. Check again, I updated it to use the actual class name which is uppercase and so this function will transform that.

easypost/services/base_service.py
Justintime50: Made them all underscored, let me know if that's not a good idea. The problem with the class inheritance we had before was it was a bit magical and didn't allow for us to be explicit in each function call which we now need to be because of the client object (need to pass in each model string, etc)

jchen293: If we don't want users to use these internal functions or just show up in their IDE auto-completion. We should be explicit for each service (that's what I am doing in Ruby) but more code for the lib. There is a trade-off decision we need to think about, would like more feedback from other people on this.

wildsparx:
The underscore solution sounds fine to me. It's a convention that you can expect the caller to respect.

easypost/services/base_service.py
nwithan8: I agree with putting them in the parent class (to keep the code DRY) and making them "internal" by marking with the leading underscore. Python as a language isn't good with access levels, but we can do our best to work within the conventions of the language, under the assumption that our end-users are respecting those. If an end-user chooses to access an underscored method and odd behavior happens as a result, that's on them at that point.

tests/test_easypost_client.py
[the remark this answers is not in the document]
nwithan8: api_version will be a variable set by each endpoint call, not during client initialization.

jchen293 (dismissed):
Underscore is the way we are going for. Only thing left is the missing type hints need to be added in the next PR

nwithan8 (approved):

jchen293 (approved):

PR #275, opened 2023-05-15
services2

Justintime50:
# Description

- feat: BetaCarrierMetadataService
- feat: BetaRateService
- feat: BetaReferralCustomerService
- feat: EndShipperService
- feat: EventService
- feat: InsuranceService
- feat: OrderService
- feat: RateService

# Testing

- Updated tests

# Pull Request Type


easypost/easypost_client.py
```diff
@@ -22,7 +30,15 @@ def __init__(self, api_key: str, api_base: str = API_BASE, timeout: int = TIMEOU
         # Services
         self.address = AddressService(self)
         self.batch = BatchService(self)
+        self.beta_carrier_metadata = BetaCarrierMetadataService(self)
```
nwithan8, line 33: Thought: Since users are only accessing this instance of the Carrier Metadata service/function group via the `beta_carrier_metadata` property, we don't have to call the class `BetaCarrierMetadata` since users only know it as `.beta_carrier_metadata`

nwithan8 (approved):

easypost/services/end_shipper_service.py
```diff
@@ -0,0 +1,49 @@
...
+
+        response, api_key = Requestor(self._client).request(method=RequestMethod.POST, url=url, params=wrapped_params)
+
+        return convert_to_easypost_object(response=response, api_key=api_key)
+
+    def all(self, **params) -> List[Address]:
+        """Retrieve a list of EndShippers."""
+        return self._all_resources(self._service_class, **params)
+
+    def retrieve(self, id) -> Address:
```
jchen293, line 35:
```suggestion
    def retrieve(self, id: str) -> Address:
```

This type hint is also missing in other services retrieve function

jchen293 (approved):

easypost/easypost_client.py
Justintime50: Until it goes GA, then we'd need both namespaces since for a time they will both exist together and logic could differ (which they will, the retrieve function name will change)

easypost/services/end_shipper_service.py
Justintime50: As mentioned elsewhere, I will come back and do type hints separately

PR #277, opened 2023-05-16
Service and Model Cleanup

Justintime50:
# Description

This PR is a follow-up cleanup after breaking out services and models and does the following:

1. Removes all old logic (combined services/models)
2. Deserializes objects into their models (to get around circular imports, we are now initializing the model at runtime instead of at import time.)
3. Uses better imports by importing in the `__init__` files for cleaner names inside the library
4. Fixes various tests that were broken and use the new syntax for all tests
5. Various other small fixes

I have another PR planned to cleanup the requestor and EasyPost object files which will follow this.

# Testing

Fixed various tests that needed that and re-recorded cassettes.

# Pull Request Type


easypost/util.py
```diff
@@ -45,7 +43,7 @@ def get_lowest_object_rate(
     return lowest_rate
 
 
-def get_lowest_smart_rate(smart_rates, delivery_days: int, delivery_accuracy: str) -> Rate:
```
Justintime50, line 48: # Reviewer Note

These type hints were removed due to circular import issues. Will try and revisit this when I come back and fix all type hints

nwithan8: Put single quotes around the word `Rate`

Justintime50: It's the import itself that is a problem. I already tried that approach

easypost/constant.py
```diff
@@ -1,6 +1,10 @@
-SUPPORT_EMAIL = "support@easypost.com"
-TIMEOUT = 60
+# flake8: noqa
+VERSION = "7.13.0"
+VERSION_INFO = [str(number) for number in VERSION.split(".")]
 API_BASE = "https://api.easypost.com/v2"
+TIMEOUT = 60
+
+SUPPORT_EMAIL = "support@easypost.com"
```
jchen293, line 7: Nit: Sort these constants

Need a timeout test for the client, or is this happening in the next PR?

easypost/util.py
```diff
@@ -99,7 +97,7 @@ def get_lowest_stateless_rate(
     return lowest_rate
 
 
-def receive_event(raw_input: str) -> Event:
+def receive_event(raw_input: str):
     """Receives a raw Webhook event and converts it to JSON."""
     # TODO: Remove api_key
```
jchen293, line 102: Question: should this TODO be part of this PR?

jchen293 (commented):
Seems fine, there are missing type hints for some parameters that need to be added but I assume there will be a follow-up PR to address that?

easypost/constant.py
Justintime50: It's still on my TODO list, will address before we release.

As for the constants, version is first because it will be the one changing most often so wanted easy access to it at the top

jchen293 (approved):
seems fine, nice clean up work!

nwithan8 (approved):

PR #278, opened 2023-05-17
Cleanup2

### LINKED ISSUE #97: ResourceWarning: Unclosed ssl.SSLSOCKET
This may or may not be an issue as from my research you sometimes want to keep SSL sockets open but I wanted to report this in the off chance we need to address it. When `PYTHONWARNINGS=always` is used, I get the following warning:

```
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=5, family=AddressFamily.AF_INET6, 
type=SocketKind.SOCK_STREAM, proto=0, laddr=('2601:681:201:15ec:5c9c:82f1:1c10:f6a9', 
51562, 0, 0), raddr=('2607:f0d0:3803:ca::3', 443, 0, 0)>
```

Justintime50:
# Description

Cleans up `easypost_object` and `requestor` by removing dead code, fixing style, and moving the requestor logic traditionally setup during import to the init of the client object finishing off thread-safety for this library.

Closes #97 

# Testing

Updated tests as needed

# Pull Request Type


easypost/easypost_client.py
```diff
@@ -33,7 +35,7 @@
 class EasyPostClient:
     """A client object used to authenticate and configure all HTTP calls to the EasyPost API."""
 
-    def __init__(self, api_key: str, api_base: str = API_BASE, timeout: int = TIMEOUT):
+    def __init__(self, api_key: str, api_base: str = f"{API_BASE}/{API_VERSION}", timeout: int = TIMEOUT):
```
jchen293, line 38: Nit: prefer to break this line of parameters

jchen293 (approved):
lgtm!

easypost/easypost_client.py
Justintime50: Sure, can do in the next one

nwithan8 (approved):
Not a huge fan of how the API version gets set (would prefer on a per-request basis rather than instance-wide, to allow for "beta" when needed without splitting strings), but won't block on it.

Justintime50:
> Not a huge fan of how the API version gets set (would prefer on a per-request basis rather than instance-wide, to allow for "beta" when needed without splitting strings), but won't block on it.

It already works that way, if you pass the `beta=True` param, we replace the version with `beta` 🙂 Unless I'm misunderstanding?

PR #279, opened 2023-05-18
Error Overhaul

Justintime50:
# Description

Added the following API errors that all extend `ApiError` which then extends `EasyPostError`:

- ApiError
- EncodingError
- ExternalApiError
- ForbiddenError
- GatewayTimeoutError
- HttpError
- InternalServerError
- InvalidRequestError
- JsonError
- MethodNotAllowedError
- NotFoundError
- PaymentError
- RateLimitError
- RedirectError
- ServiceUnavailableError
- TimeoutError
- UnauthorizedError
- UnknownApiError

Added the following general errors that extend `EasyPostError` which then extends `Exception`:

- EasyPostError
- EndOfPaginationError
- FilteringError
- InvalidObjectError
- InvalidParameterError
- MissingParameterError
- SignatureVerificationError

These errors are either thrown directly or thrown as a part of handling the response from the API.

# Testing

Tests were updated to accomadate new error properties.

# Pull Request Type


easypost/requestor.py
```diff
@@ -16,11 +16,47 @@
...
+    TimeoutError,
+    UnauthorizedError,
+    UnknownApiError,
+)
+
+
+STATUS_CODE_TO_ERROR_MAPPING: Dict[int, Any] = {
+    401: UnauthorizedError,
+    402: PaymentError,
+    403: UnauthorizedError,
```
jchen293, line 50: This should be `ForbiddenError`

tests/conftest.py
```diff
@@ -18,7 +18,6 @@
 EASYPOST_PROD_API_KEY = os.getenv("EASYPOST_PROD_API_KEY")
 PARTNER_USER_PROD_API_KEY = os.getenv("PARTNER_USER_PROD_API_KEY", "123")
 REFERRAL_CUSTOMER_PROD_API_KEY = os.getenv("REFERRAL_CUSTOMER_PROD_API_KEY", "123")
-STRIPE_CONNECT_USER_PROD_API_KEY = os.getenv("STRIPE_CONNECT_USER_PROD_API_KEY", "123")
```
jchen293, line 21: Question: why was this removed?

jchen293 (changes requested):
Overall seems fine, left one blocking change, and also need a test for `TimeoutError`, I believe there are left TODOs for the `easypost_client` test IIRC.

tests/conftest.py
Justintime50: It was added previously when I was fixing the partner tests but wasn't needed after all. The fixture was removed earlier but the constant remained and is unused so I removed it.

easypost/requestor.py
Justintime50: Nice catch! Fixed

jchen293 (approved):

nwithan8 (approved):

PR #287, opened 2023-06-13
[chore] Import style guide from examples submodule

nwithan8:
# Description

- Reference the style guides from the examples submodule when formatting/linting (thankfully, no symlinking needed)
- Update Make steps, CI as needed to account for style guide imports
- Apply suggested linting/formatting as needed

# Testing

- Make steps work as expected locally

# Pull Request Type


easypost/requestor.py
```diff
@@ -15,7 +15,6 @@
 from urllib.parse import urlencode
 
 import requests
-
```
Justintime50, line 18: issue: the fact that all of these isort changes were made indicates that the config isn't getting picked up correctly. There should be no changes needed. Let's investigate what we need to do to get it to where isort doesn't need to make changes in this PR

Justintime50 (changes requested):

easypost/requestor.py
nwithan8: It seems to be due to a discrepancy between ["FIRSTPARTY" and "LOCALFOLDER"](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html) when the style guide is referenced in a non-root capacity.

Adding `no_lines_before="LOCALFOLDER"` no longer causes these files to be touched in the client library, and has no effect on Python code in the examples repo. We can choose to adopt this minor alteration to the style guide in the examples repo, or we can leave it as is and apply the rules as interpreted to the client library.

Justintime50: What's your suggestion?

nwithan8: I'd be in favor of keeping the style as is and accepting the linting. From what I can tell, it should have been linting it like this all along, since pytest is in a different category than the local `easypost` files.

Justintime50 (approved):

jchen293 (approved):

PR #300, opened 2023-10-20
[feat] Add beta feature to retrieve paginated child user list

nwithan8:
# Description

- Add `easypost.beta.User.retrieve_all_children` and `easypost.beta.User.get_next_page_of_children` functions to retrieve, iterate over paginated list of child users.

Could not use existing base classes due to incompatibilities with the "class URL"

**NOTE: This is being backported to v7.**

# Testing

- Add unit tests, cassettes as needed

# Pull Request Type


dcaballeroc (approved):
Looks legit

easypost/beta/user.py
```diff
@@ -0,0 +1,51 @@
...
+    RequestMethod,
+    Requestor,
+)
+from easypost.resource import Resource
+
+
+class User(Resource):
+    @classmethod
+    def retrieve_all_children(cls, api_key: Optional[str] = None, **params) -> Dict[str, Any]:
+        """Retrieve a list of children from the API."""
```
jchen293, line 20:
```suggestion
        """Retrieve a list of paginated children from the API."""
```

easypost/beta/user.py
```diff
@@ -0,0 +1,51 @@
...
+        children: Dict[str, Any],
+        page_size: int,
+        api_key: Optional[str] = None,
+    ) -> List["User"]:
+        """Get next page of children."""
+        requestor = Requestor(local_api_key=api_key)
+        url = "/users/children"
+        children_array = children.get("children", [])
+
+        if children_array is None or len(children_array) == 0 or not children.get("has_more"):
```
jchen293, line 38:
```suggestion
        if children_array is None or not children.get("has_more"):
```

Let's remove the redundant check

easypost/beta/user.py
```diff
@@ -0,0 +1,51 @@
...
+            raise Error(message="There are no more pages to retrieve.")
+
+        params = {
+            "after_id": children_array[-1].id,
+            "page_size": page_size,
+        }
+
+        response, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
+        response_array: List[Any] = response.get(url[1:])  # type: ignore
+        if response is None or len(response_array) == 0 or not response.get("has_more"):
```
jchen293, line 48: same here

easypost/beta/user.py
```diff
@@ -0,0 +1,51 @@
...
+        if children_array is None or len(children_array) == 0 or not children.get("has_more"):
+            raise Error(message="There are no more pages to retrieve.")
+
+        params = {
+            "after_id": children_array[-1].id,
+            "page_size": page_size,
+        }
+
+        response, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
+        response_array: List[Any] = response.get(url[1:])  # type: ignore
```
jchen293, line 47: what is the type ignore here for?

easypost/beta/user.py
Justintime50: +1, calling this out is important as it's the differentiating factor of this new endpoint/function and defining this well will lead to greater adoption for people looking for a solution to the current problem of all children being too large for some users

nwithan8: For the "retrieve all" methods, we typically call it `.all()`, but that would get confused with "retrieve all users", so we could do `all_children()`?

nwithan8: Resolved

easypost/beta/user.py
nwithan8: Good catch, revealed a bug

easypost/beta/user.py
```diff
@@ -0,0 +1,51 @@
...
+        if len(children_array) == 0 or not children.get("has_more", False):
+            raise Error(message="There are no more pages to retrieve.")
+
+        params = {
+            "after_id": children_array[-1].id,
+            "page_size": page_size,
+        }
+
+        data, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
+        next_children_array: List[Any] = data.get("children", [])
```
jchen293, line 47: The result is coming back from an API call, `data` should be called `response` like what you had

```suggestion
        response, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
        next_children_array: List[Any] = response.get("children", [])
```

nwithan8: But it's not actually the HTTP response, but the JSON data. Either way, it doesn't affect the functionality, just clarifies the internal variable names.

jchen293: Alright, if it works then it works
image

jchen293 (approved):

easypost/beta/user.py
Justintime50: I agree that response is probably a more appropriate name for this variable

Justintime50 (approved):

PR #303, opened 2023-10-30
[fix] Remove bad post-request guard clause in pagination functions

nwithan8:
# Description

The highlighted code (removed as part of this PR) was an invalid guard clause added at the root of our pagination functionality (and re-implemented for a recent beta child user pagination function in #300 ).

image

The highlighted code checked AFTER retrieving the "next page" of results if there were still more results to retrieve (the `has_more` boolean). If this was `False`, an error would be thrown alerting the user that there were no more pages to retrieve. However, in doing so, this would discard what we would then know is the last valid page of results, never returning them to the end-user. In effect, the end-user was never able to collect the very last page of results, because the function was always prematurely checking if yet another page exists afterwards. (The array length check also included in this guard clause is largely inconsequential, as it should never be triggered; it more so validates the shape of the response data, catching issues with the current page of results).

This PR removes the invalid guard clause both in the base class and the one-off re-implementation in the beta User service, and adds a unit test for the beta User service that validates the functionality. The unit test confirms that the pagination function no longer prematurely looks ahead and does in fact allow users to collect all available pages.

**This PR is for v7, and will need to be ported to v8 as well.**

# Testing

- Unit test confirms valid functionality (passes; re-add the bad line and it fails as expected)
- Added cassette (not used by test due to mocked responses, but prevents making bogus live API calls)

# Pull Request Type

### DIFF
--- a/easypost/beta/user.py
+++ b/easypost/beta/user.py
@@ -45,7 +45,7 @@ class User(Resource):
 
         data, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
         next_children_array: List[Any] = data.get("children", [])
-        if len(next_children_array) == 0 or not data.get("has_more", False):
+        if len(next_children_array) == 0:
             raise Error(message="There are no more pages to retrieve.")
 
         return convert_to_easypost_object(response=data, api_key=api_key)
--- a/easypost/resource.py
+++ b/easypost/resource.py
@@ -93,7 +93,7 @@ class NextPageResource(Resource):
 
         response, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params)
         response_array: List[Any] = response.get(url[1:])  # type: ignore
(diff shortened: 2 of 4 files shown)


easypost/beta/user.py
```diff
@@ -45,7 +45,7 @@ def get_next_page_of_children(
 
         data, api_key = requestor.request(method=RequestMethod.GET, url=url, params=params, beta=True)
         next_children_array: List[Any] = data.get("children", [])
-        if len(next_children_array) == 0 or not data.get("has_more", False):
+        if len(next_children_array) == 0:
```
Justintime50, line 48: question: should this have just been an `and` instead of an `or`?

Justintime50 (approved):

easypost/beta/user.py
jchen293: +1

jchen293 (approved):

PR #307, opened 2023-10-31
[fix] Fix, re-design implementation of pagination functions

nwithan8:
- Redesign pagination for more explicit per-service implementation, support for filters, less fragile

# Description

As we've recently discovered, a bug in the implementation of the "get_next_page" functions in this library meant any specific filters used while getting pages for a given resource were not being preserved and carried forward (e.g. the type of report, a specific tracking code for trackers, "purchased" filter for shipments).

This PR redesigned the underlying implementation of the "get_next_page" functions to be more explicit and account for minor inconsistencies endpoint-to-endpoint by relying on a hidden "_filters" key that will carry necessary metadata about parameters and filters used for each "all" and "get_next_page" call for easy reference behind the scenes when retrieving the next page of a paginated collection. From an end-user perspective, this means they will not need to always pass in, e.g., the "purchased" filter for each page request of shipments, or the type of report for each page request of reports. Whatever filters were used to retrieve the first page of a paginated collection will be reused automatically behind-the-scenes when retrieving each subsequent page. ("page_size" is still on a per-page basis, and filters can be overridden with optional parameters if necessary). This more closely follows the design of pagination as implemented in the Java and .NET library, which were the original models for these functions.

This PR does not change the majority of the end-user experience or public function signatures. As demonstrated by the minor changes to the unit tests and the lack of need to re-record any cassettes, this implementation change does not alter the shape or workflow as it previously was.

The only "breaking changes" include:
- "get_next_page" for the Batch service now throws a `NotImplementedError`, as there is a known bug with batch pagination server-side, and this function should not have been available in the first place.
- "get_next_page" for the Order service now throws a `NotImplementedError`, as order pagination does not exist server-side, and this function should not have been available in the first place. There was not an "all" function for Orders in the first place, so it's unclear how an end-user would have been able to retrieve the first page to use with this function anyhow.

# Testing

- All pagination unit tests have had assertions added to verify the new "filters" are being passed forward as expected, and that the implementation changes do not alter the existing end-user experience
- Only the "referral customer get next page" cassette needed to be re-recorded, as new server-side data allowed for a second page and therefore a second HTTP request/response to be recorded.

# Pull Request Type

### DIFF
--- a/easypost/services/base_service.py
+++ b/easypost/services/base_service.py
@@ -6,7 +6,10 @@ from typing import (
     Optional,
 )
 
-from easypost.constant import NO_MORE_PAGES_ERROR
+from easypost.constant import (
+    NO_MORE_PAGES_ERROR,
+    _FILTERS_KEY,
+)
 from easypost.easypost_object import convert_to_easypost_object
 from easypost.errors import EndOfPaginationError
 from easypost.requestor import (
@@ -46,11 +49,14 @@ class BaseService:
 
         return convert_to_easypost_object(response=response)
 
-    def _all_resources(self, class_name: str, **params) -> Any:
+    def _all_resources(self, class_name: str, filters: Dict[str, Any] = None, **params) -> Any:
         """Retrieve a list of EasyPostObjects from the EasyPost API."""
         url = self._class_url(class_name)
-
         response = Requestor(self._client).request(method=RequestMethod.GET, url=url, params=params)
+
+        if filters: # presence of filters indicates we are dealing with a paginated response
+            self._check_has_current_page(response=response, filters=filters)
+            response[_FILTERS_KEY] = filters # Save the filters used to reference in potential get_next_page call
 
         return convert_to_easypost_object(response=response)
 
@@ -79,32 +85,20 @@ class BaseService:
 
         return convert_to_easypost_object(response=response)
 
-    def _get_next_page_resources(
-        self,
-        class_name: str,
-        collection: Dict[str, Any],
-        page_size: int,
-        optional_params: Optional[Dict[str, Any]] = None,
-    ) -> Dict[str, Any]:
-        """Retrieve next page of EasyPostObjects via the EasyPost API."""
-        url = self._class_url(class_name)
-        collection_array = collection.get(url[1:])
-
-        if collection_array is None or len(collection_array) == 0 or not collection.get("has_more"):
+    def _check_has_next_page(self, collection: Dict[str, Any]) -> None:
+        """Raise exception if there is no next page of a collection."""
+        if not collection.get("has_more", False):
             raise EndOfPaginationError(NO_MORE_PAGES_ERROR)
 
-        params = {
-            "before_id": collection_array[-1].id,
-            "page_size": page_size,
-        }
-
-        if optional_params:
-            params.update(optional_params)
-
-        response = Requestor(self._client).request(method=RequestMethod.GET, url=url, params=params)
-
-        response_array: List[Any] = response.get(url[1:])  # type: ignore
-        if response is None or len(response_array) == 0:
+    def _check_has_current_page(self, response: Dict[str, Any], filters: Dict[str, Any]) -> None:
+        """Raise exception if there is no current page of a collection."""
+        if not response:
             raise EndOfPaginationError(NO_MORE_PAGES_ERROR)
 
-        return convert_to_easypost_object(response=response)
+        entries_key = filters.get("key", None)
+        if not entries_key:
+            raise EndOfPaginationError(NO_MORE_PAGES_ERROR)
+
+        entries = response.get(entries_key, [])
+        if not entries or len(entries) == 0:
+            raise EndOfPaginationError(NO_MORE_PAGES_ERROR)
--- a/easypost/services/batch_service.py
+++ b/easypost/services/batch_service.py
@@ -24,7 +24,11 @@ class BatchService(BaseService):
 
     def all(self, **params) -> Dict[str, Any]:
         """Retrieve a list of Batches."""
-        return self._all_resources(self._model_class, **params)
+        filters = {
+            "key": "batches",
+        }
+
+        return self._all_resources(self._model_class, filters, **params)
 
     def retrieve(self, id: str) -> Batch:
         """Retrieve a Batch."""
@@ -85,5 +89,10 @@ class BatchService(BaseService):
         page_size: int,
         optional_params: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
-        """Retrieve the next page of the list Batch response."""
-        return self._get_next_page_resources(self._model_class, batches, page_size, optional_params)
+        """
+        Retrieve the next page of the list Batch response.
+
+        NOTE: This function has known issues and should not be used.
+        """
+        # API doesn't return batches newest to oldest, so these parameters don't work as expected
+        raise NotImplementedError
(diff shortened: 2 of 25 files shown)


easypost/services/batch_service.py
```diff
@@ -85,5 +89,10 @@ def get_next_page(
...
     ) -> Dict[str, Any]:
-        """Retrieve the next page of the list Batch response."""
-        return self._get_next_page_resources(self._model_class, batches, page_size, optional_params)
+        """
+        Retrieve the next page of the list Batch response.
+
+        NOTE: This function has known issues and should not be used.
+        """
+        # API doesn't return batches newest to oldest, so these parameters don't work as expected
+        raise NotImplementedError
```
Justintime50, line 98: issue: the docstring is good, but we can't just rip this out. Even if the params don't work as anticipated, this is a bug with the API and not this lib. If someone was using the lib today without the params, this would do just fine. To now raise a `NotImplementedError` is a breaking change and should not happen without a major release.

easypost/services/event_service.py
```diff
@@ -56,4 +60,14 @@ def get_next_page(
         optional_params: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         """Retrieve the next page of the list Events response."""
-        return self._get_next_page_resources(self._model_class, events, page_size, optional_params)
+        self._check_has_next_page(collection=events)
```
Justintime50, line 63: question: does it make sense to put all of this into the base class so that every service can reuse it and we only override those we need (reports, shipments, etc)? That way we save on code reuse. Wouldn't that get us the best of both worlds?

easypost/services/order_service.py
```diff
@@ -28,12 +28,12 @@ def retrieve(self, id: str) -> Order:
...
     def get_next_page(
         self,
-        insurances: Dict[str, Any],
+        orders: Dict[str, Any],
         page_size: int,
         optional_params: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         """Retrieve the next page of the list Order response."""
-        return self._get_next_page_resources(self._model_class, insurances, page_size, optional_params)
+        raise NotImplementedError
```
Justintime50, line 36: gosh... how did this make it in initially? 😭

Do we want to not implement this or just remove it? It shouldn't have been here at all so I'm less worried about removal here, especially since we don't have an `all` endpoint to pass orders into - and it was mislabeled insurances 😂

easypost/services/report_service.py
```diff
@@ -35,15 +38,21 @@ def create(self, **params) -> Report:
...
+            "key": "reports",
+            "type": params.get("type", None),
+        }
 
-        if refund_type is None:
+        report_type = params.pop("type")
+        if report_type is None:
             raise MissingParameterError(MISSING_PARAMETER_ERROR.format("type"))
 
-        url = f"{self._class_url(self._model_class)}/{refund_type}"
```
Justintime50, line 43: these variable typos pain me

Justintime50 (dismissed):

easypost/services/shipment_service.py
```diff
@@ -54,12 +57,23 @@ def get_next_page(
...
+        self._check_has_next_page(collection=shipments)
+
+        params = {
+            "before_id": shipments["shipments"][-1].id,
+            "page_size": page_size,
+            "include_children": shipments.get(_FILTERS_KEY, {}).get(
+                "include_children", None
+            ),  # Use the same include_children as the last page
+            "purchased": shipments.get(_FILTERS_KEY, {}).get(
+                "purchased", None
```
jchen293, line 69:
```suggestion
                "include_children"
            ),  # Use the same include_children as the last page
            "purchased": shipments.get(_FILTERS_KEY, {}).get(
                "purchased"
```
`None` is already the default, let's not include it

easypost/services/tracker_service.py
```diff
@@ -44,12 +44,21 @@ def get_next_page(
...
-            "carrier": trackers.get("carrier"),
+        self._check_has_next_page(collection=trackers)
+
+        params = {
+            "before_id": trackers["trackers"][-1].id,
+            "page_size": page_size,
+            "tracking_code": trackers.get(_FILTERS_KEY, {}).get(
+                "tracking_code", None
+            ),  # Use the same tracking_code as the last page
+            "carrier": trackers.get(_FILTERS_KEY, {}).get("carrier", None),  # Use the same carrier as the last page
```
jchen293, line 55: Same here, let's not include `None`

easypost/services/batch_service.py
nwithan8: Resolved

easypost/services/event_service.py
nwithan8: I've moved as much generic code to the base class that I feel comfortable with. The code remaining as per-service implementations is because there are/will be variance per-service (e.g. batches I know needs to be "after_id", not "before_id", the key to look for in each JSON array differs).

easypost/services/order_service.py
nwithan8: I'd be down for just ripping it out.

easypost/services/base_service.py
```diff
@@ -46,12 +47,14 @@ def _create_resource(self, class_name: str, **params) -> Any:
 
         return convert_to_easypost_object(response=response)
 
-    def _all_resources(self, class_name: str, **params) -> Any:
+    def _all_resources(self, class_name: str, filters: Dict[str, Any] = None, **params) -> Any:  # type: ignore
```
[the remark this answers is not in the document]
nwithan8: Doesn't like that the `filters` has a default "None" when it's supposed to be type `Dict[str, Any]`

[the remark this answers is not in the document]
nwithan8: Resolved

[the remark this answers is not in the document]
Justintime50: Why not have it as an empty dict and update whatever checks are needed. Empty dict is falsy so we could get away with that while satisfying the type checker

easypost/services/shipment_service.py
nwithan8: Resolved

easypost/services/order_service.py
Justintime50: Let's rip it out then.

@jchen293 because it shouldn't have ever made it in here because there's no `all` function for orders, you can't pass in the results anyway to get the next page.

easypost/services/tracker_service.py
nwithan8: Resolved

easypost/services/batch_service.py
```diff
@@ -95,4 +95,14 @@ def get_next_page(
         NOTE: This function has known issues and should not be used.
```
Justintime50, line 95: please specify what the known issues are - they are known to us but maybe not to users, let's define it here so they can make an educated decision.

easypost/services/shipment_service.py
```diff
@@ -62,12 +62,9 @@ def get_next_page(
...
             "before_id": shipments["shipments"][-1].id,
             "page_size": page_size,
-            "include_children": shipments.get(_FILTERS_KEY, {}).get(
-                "include_children", None
-            ),  # Use the same include_children as the last page
-            "purchased": shipments.get(_FILTERS_KEY, {}).get(
-                "purchased", None
-            ),  # Use the same purchased as the last page
+            "include_children": shipments.get(_FILTERS_KEY, {}).get("include_children"),
+            # Use the same include_children as the last page
```
Justintime50, line 66: this comment should be above what it's referencing if it's unable to be on the same line

easypost/services/batch_service.py
nwithan8: Resolved

Justintime50 (approved):

jchen293 (approved):

PR #318, opened 2023-12-19
fix regenerate rates type

### LINKED ISSUE #317: [Bug]: `ShipmentService.regenerate_rates` has incorrect return value type
### Software Version

v9.0.0

### Language Version

3.12

### Operating System

Fedora

### What happened?

The return value type of `ShipmentService.regenerate_rates` is incorrect. It is typed as a `Shipment`, but the Easypost API returns a payload of

```json
{
  "rates": [
    ...
  ]
}
```

### What was expected?

The return value should be correctly typed to reflect the rates payload returned by the API.

### Sample Code

_No response_

### Relevant logs

_No response_

Justintime50:
# Description

Fixes the return type of the `regenerate_rates` function. Closes #317 

# Testing

# Pull Request Type

### DIFF
--- a/easypost/services/shipment_service.py
+++ b/easypost/services/shipment_service.py
@@ -72,7 +72,7 @@ class ShipmentService(BaseService):
 
         return self.all(**params)
 
-    def regenerate_rates(self, id: str) -> Shipment:
+    def regenerate_rates(self, id: str) -> Dict[str, List[Rate]]:
         """Regenerate Rates for a Shipment."""
         url = f"{self._instance_url(self._model_class, id)}/rerate"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v9.0.1 (2023-12-20)
+
+- Corrects the return type of `regenerate_rates`
+
 ## v9.0.0 (2023-12-06)
 
 See our [Upgrade Guide](UPGRADE_GUIDE.md#upgrading-from-8x-to-90) for more details.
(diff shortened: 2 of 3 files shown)


nwithan8 (dismissed):

easypost/services/shipment_service.py
```diff
@@ -72,7 +72,7 @@ def get_next_page(
 
         return self.all(**params)
 
-    def regenerate_rates(self, id: str) -> Shipment:
+    def regenerate_rates(self, id: str) -> Dict[str, List[Rate]]:
```
bradydean, line 75: While this works because of duck typing, it would be better from a client perspective if the return type inherited from `EasyPostObject` since the method does return an `EasyPostObject` instance.

In my usage of this method, I pass the return value to `util.get_lowest_object_rate`, and with this `Dict` type it would fail to typecheck with pyright.

jchen293 (dismissed):

jchen293 (approved):

easypost/services/shipment_service.py
Justintime50: Thanks for clarifying your use case! The `get_lowest_object_rate` function is intended for internal use by our other `get_lowest_rate` functions on a per-object basis, though it could certainly be used by end-users. The proposed return value is definitely the correct choice here; however, I can appreciate the desire to pass in the regenerated rates to the `lowest_rate` function. I've loosened the typing on `get_lowest_object_rate` to allow bare dictionaries in addition to EasyPostObjects which will retain the intention that this is for EasyPostObjects while allowing you to use it in the way you mentioned. We'll get this released in `v9.0.1`, please let us know if this doesn't fix the issue!

PR #329, opened 2024-03-28
Add Refund function in insurance service

jchen293:
# Description

Add Refund function in insurance service for insurance cancellation endpoint

# Testing

Add unit test

# Pull Request Type


pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+testpaths=tests
```
nwithan8, line 2: A doc generation tool introduced into our examples submodule a few months ago uses pytest and its own `conftest.py` file to execute.

When trying to run the unit tests on this project, pytest was discovering that `conftest.py` as well as the one for this test suite and complaining about multiple conf files.

This `pytest.ini` configuration files tests pytest to ONLY look in the `tests` folder when executing the test suite (via our `make test` step).

nwithan8 (dismissed):

pytest.ini
jchen293: Thanks for adding the note to it!

easypost/services/insurance_service.py
```diff
@@ -47,3 +52,13 @@ def get_next_page(
             params.update(optional_params)
 
         return self.all(**params)
+
+    def refund(self, id: str) -> Insurance:
+        url = f"/insurances/{id}/refund"
+        response = Requestor(self._client).request(
+            method=RequestMethod.POST,
+            url=url,
+            beta=True,  # Remove beta when endpoint is in GA.
```
Justintime50, line 61:
```suggestion
            beta=True,  # TODO: Remove beta when endpoint is in GA.
```

Because this requires follow-up. we should add `TODO` to ensure we differentiate this from a normal code comment meant to explain something

Justintime50 (approved):

easypost/services/insurance_service.py
jchen293: Good catch, I forgot to add it. This diff will not be landed until the endpoint is in GA. Marking this PR back to draft

Justintime50 (dismissed):

nwithan8 (approved):

nwithan8 (approved):

PR #336, opened 2024-07-03
feat: add new create_ups and update_ups functions

Justintime50:
# Description

- Routes `UpsAccount`, `UpsMailInnovationsAccount`, and `UpsSurepostAccount` create/update requests to the new `/ups_oauth_registrations` endpoint

# Testing

Adds new tests for each, updates previous custom workflow test to use FedEx instead of UPS

# Pull Request Type

### DIFF
--- a/easypost/services/base_service.py
+++ b/easypost/services/base_service.py
@@ -6,6 +6,7 @@ from typing import (
 )
 
 from easypost.constant import (
+    _EXCLUDED_CLASS_NAMES,
     _FILTERS_KEY,
     NO_MORE_PAGES_ERROR,
 )
@@ -30,7 +31,9 @@ class BaseService:
     def _class_url(self, class_name: str) -> str:
         """Generate a URL based on class name."""
         transformed_class_name = self._snakecase_name(class_name)
-        if transformed_class_name[-1:] in ("s", "h"):
+        if transformed_class_name in _EXCLUDED_CLASS_NAMES:
+            return f"/{transformed_class_name}"
+        elif transformed_class_name[-1:] in ("s", "h"):
             return f"/{transformed_class_name}es"
         else:
             return f"/{transformed_class_name}s"
--- a/easypost/services/carrier_account_service.py
+++ b/easypost/services/carrier_account_service.py
@@ -7,6 +7,7 @@ from typing import (
 
 from easypost.constant import (
     _CARRIER_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS,
+    _UPS_OATH_CARRIER_ACCOUNT_TYPES,
     MISSING_PARAMETER_ERROR,
 )
 from easypost.easypost_object import convert_to_easypost_object
@@ -32,7 +33,10 @@ class CarrierAccountService(BaseService):
             raise MissingParameterError(MISSING_PARAMETER_ERROR.format("type"))
 
         url = self._select_carrier_account_creation_endpoint(carrier_account_type=carrier_account_type)
-        wrapped_params = {self._snakecase_name(self._model_class): params}
+        if carrier_account_type in _UPS_OATH_CARRIER_ACCOUNT_TYPES:
+            wrapped_params = {"ups_oauth_registrations": params}
+        else:
+            wrapped_params = {self._snakecase_name(self._model_class): params}
 
         response = Requestor(self._client).request(method=RequestMethod.POST, url=url, params=wrapped_params)
 
@@ -48,7 +52,11 @@ class CarrierAccountService(BaseService):
 
     def update(self, id: str, **params) -> CarrierAccount:
         """Update a CarrierAccount."""
-        return self._update_resource(self._model_class, id, **params)
+        if params.get("type") in _UPS_OATH_CARRIER_ACCOUNT_TYPES:
+            class_name = "UpsOauthRegistrations"
+        else:
+            class_name = self._model_class
+        return self._update_resource(class_name, id, **params)
 
     def delete(self, id: str) -> None:
         """Delete a CarrierAccount."""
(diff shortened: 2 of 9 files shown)


jchen293 (approved):
lgtm

nwithan8 (approved):

tests/test_carrier_account.py
```diff
@@ -93,3 +93,41 @@ def test_carrier_account_create_with_custom_workflow(prod_client):
...
+    which we don't have in a test context, simply assert that we routed the request correctly.
+    """
+    carrier_account = {
+        "type": "UpsAccount",
+    }
+
+    try:
+        prod_client.carrier_account.update_ups("ca_123", **carrier_account)
+    except ApiError as error:
+        assert error.http_status == 404
```
jchen293, line 133: I just revisited this after finishing the Java library. We should try to avoid asserting a 404 error here because it might not be hitting the correct endpoint. Instead, you can reuse the test from the update above and simply pass in bad data to get a 422 error from the UPS update endpoint. This will give us more confidence that we are actually hitting the correct endpoint, as we can't truly test this.

Justintime50: Actually double checking this, the reason I had it as 404 is we can't create a valid UPS account so there wouldn't be one to update. That's why the best we can do is 404 which isn't great but then we can look at the cassettes to make sure it's hitting the right place. Mind taking a look again?

easypost/services/carrier_account_service.py
```diff
@@ -48,7 +52,11 @@ def retrieve(self, id: str) -> CarrierAccount:
 
     def update(self, id: str, **params) -> CarrierAccount:
         """Update a CarrierAccount."""
-        return self._update_resource(self._model_class, id, **params)
+        if params.get("type") in _UPS_OATH_CARRIER_ACCOUNT_TYPES:
```
jchen293, line 55: I am not sure if this will work. `Type` isn't a required field for update and we can't rely on it: https://docs.easypost.com/docs/carrier-accounts#update-a-carrieraccount

tests/test_carrier_account.py
jchen293: You don't need to create a UPS account. Instead, you can reuse the [carrier create](https://github.com/EasyPost/easypost-python/blob/master/tests/test_carrier_account.py#L19) and then pass the carrier ID into the UPS create endpoint. This approach will result in a 422 error, which is much better than receiving a 404. Without looking at the cassette, it would be difficult for future readers to determine whether the 404 occurred due to an incorrect URL endpoint or an invalid carrier ID.

Justintime50: This only works today because they haven't yet turned off the old method of creating a UPS account. In a couple weeks, this will stop working so we should not rely on it.

easypost/services/carrier_account_service.py
Justintime50: Indeed. We'd need to retrieve it first then and pull out the `type` param. I don't love it but it'll make it easy on the users at the cost of a few more ms of time.

tests/test_carrier_account.py
jchen293: We are creating `DhlEcsAccount`: https://github.com/EasyPost/examples/blob/master/official/fixtures/client-library-fixtures.json#L37

tests/test_carrier_account.py
```diff
@@ -118,11 +120,14 @@ def test_carrier_account_create_ups(prod_client):
 
 
 @pytest.mark.vcr()
-def test_carrier_account_update_ups(prod_client):
+@patch("easypost.services.carrier_account_service.CarrierAccountService.retrieve")
```
jchen293, line 123: If you look at the cassette, it's using the `/carrier_accounts` endpoint for updates because this is being mocked. We should try to assert 422 down below like what I suggest above.

Justintime50: Check again, I forgot to update the return type. Mocking this works as expected. If you look above, I responded as to why using a 422 will not work anymore in a couple weeks and we cannot rely on it.

jchen293: We are creating `DhlEcsAccount` in the test, not UPS, so it should still work in a couple of weeks unless I missed something. Anyway, it’s not a dealbreaker, just something I think we can improve.

jchen293 (approved):

tests/test_carrier_account.py
Justintime50: Yeah I feel like creating a DHL account then attempting to update it view the UPS flow is even more confusing and a worse test because it's mixing things that aren't compatible. I like the thought to get it to 422, I just don't think this particular approach is the right way to do it.

nwithan8 (dismissed):

tests/test_carrier_account.py
```diff
@@ -99,45 +99,35 @@ def test_carrier_account_create_with_custom_workflow(prod_client):
...
+        "account_number": "123456789",
     }
 
-    try:
-        prod_client.carrier_account.update("ca_123", **carrier_account)
-    except ApiError as error:
-        assert error.http_status == 404
+    carrier_account = prod_client.carrier_account.create(**params)
+
+    updated_carrier_account = prod_client.carrier_account.update(carrier_account.id, account_number="987654321")
```
jchen293: From the API docs, I don't think `account_number` is a field we can update. I am also not seeing the `account_number` field in the API response. Can we update another field and then assert it down below?

easypost/services/base_service.py
```diff
@@ -30,7 +31,9 @@ def _snakecase_name(self, class_name: str) -> str:
     def _class_url(self, class_name: str) -> str:
         """Generate a URL based on class name."""
         transformed_class_name = self._snakecase_name(class_name)
-        if transformed_class_name[-1:] in ("s", "h"):
+        if transformed_class_name in _EXCLUDED_CLASS_NAMES:
+            return f"/{transformed_class_name}"
```
jchen293, line 35: What is this for? I don't quite understand why we are adding this

tests/test_carrier_account.py
Justintime50: We cannot as this updates this behind the scenes. Based on @nwithan8, this is the only (as of now) updatable field for UPS. It won't have the traditional credentials structure like other carrier accounts because it uses OAuth.

easypost/services/base_service.py
Justintime50: This is an unfortunate hack needed to get around our "magic" class transformations. Without this change, it'll double pluralize the name to `ups_oauth_registrationses` which then won't work properly. So we basically have an override list that just takes this as is without transforming it.

jchen293 (approved):
lgtm, questions answered

easypost/services/base_service.py
nwithan8: Ugh.

nwithan8 (approved):

PR #340, opened 2024-07-11
Change desired_delivery_date to str instead of key-word args

jchen293:
# Description

Change desired_delivery_date to str instead of key-word args in Shipment.recommend_ship_date function
# Testing

We should expect users to input a string and not a dict for the Shipment.recommend_ship_date function, this matches `retrieve_estimated_delivery_date` function

# Pull Request Type

### DIFF
--- a/easypost/services/shipment_service.py
+++ b/easypost/services/shipment_service.py
@@ -154,12 +154,12 @@ class ShipmentService(BaseService):
 
         return convert_to_easypost_object(response=response.get("rates", []))
 
-    def recommend_ship_date(self, id: str, **params) -> List[Dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


Justintime50 (commented):
Are we doing this for all the libs? What if a param gets added?

jchen293:
> Are we doing this for all the libs? What if a param gets added?

We haven't implemented this in the other six libraries yet, so we can make a decision here. This change matches `retrieve_estimated_delivery_date` above, which expects a string, an inconsistency I just discovered. The `precision_shipping` endpoint expects `desired_delivery_date` which should always be a single date parameter, so I don't foresee the API making a breaking change. Since both shipment smartrate endpoints are somewhat similar, this is more of a question of whether we want to wrap the parameter for users in the client libs.

Justintime50 (approved):

PR #342, opened 2024-07-19
[feat] Add claims API service

nwithan8:
# Description

Add `claim` service:
- Create a claim
- Retrieve a claim
- List all claims (paginated)
- Get the next page of claims
- Cancel a claim

# Testing

- Added unit tests and cassettes

# Pull Request Type


easypost/services/claim_service.py
```diff
@@ -0,0 +1,58 @@
...
+        self._client = client
+        self._model_class = Claim.__name__
+
+    def create(self, **params) -> Claim:
+        """Create a Claim."""
+        url = "/claims"
+
+        response = Requestor(self._client).request(method=RequestMethod.POST, url=url, params=params, beta=True)
+
+        return convert_to_easypost_object(response=response)
```
jchen293, line 27: Can we use `_create_resource` from the parent class?

nwithan8: Unfortunately not, because `_create_resource` would wrap the parameters in a top-level key, therefore sending the wrong payload schema.

jchen293: Gotcha. I will approve when endpoint is GA

jchen293 (approved):

PR #345, opened 2024-07-29
feat: automated releasing via GitHub Actions

Justintime50:
# Description

Now that we have an API Key saved to GitHub Actions for PYPI, we can use it to automate releasing this lib. Once a GitHub Release has been created (which automatically creates an accompanying tag), GitHub Actions will build and publish the library to PyPi.

# Testing

I've used this workflow in ~40 other repos for the last ~3 years with 0 problems.

# Pull Request Type

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@ name: release
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,8 @@
 name: release
 
 on:
-  push:
-    tags:
-      - '*'
+  release:
+    types: [published]
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


nwithan8 (approved):

jchen293 (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
+name: release
+
+on:
+  push:
```
nwithan8, line 4: We'll want on release, not on tag. Don't want to trigger a release when we cut, e.g. a `v7` legacy tag when we start v8 development: https://github.com/nwithan8/pytulli/blob/master/.github/workflows/publish.yml#L7

nwithan8 (changes requested):

.github/workflows/release.yml
nwithan8: I think we should also have a manual workflow dispatch as well, just in case something goes wrong we can always manually intervene.

Justintime50: Ah indeed, nice catch on the old branches.

Workflow dispatch added.

nwithan8 (approved):

nwithan8 (approved):

PR #352, opened 2024-10-10
chore: test on python 3.13

Justintime50:
# Description

1. Tests against Python 3.13
1. Install types dynamically with mypy, this is more inline with what we do internally (removes explicitly defined ones from setup.py)
1. Removes `twine` since we now use GitHub Actions to release, that package is not needed since it's already contained in the action we use

# Testing

# Pull Request Type

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,9 +20,6 @@ DEV_REQUIREMENTS = [
     "pytest-cov==4.*",
     "pytest-vcr==1.*",
     "pytest==7.*",
-    "twine==5.*;python_version>='3.8'",  # TODO: remove python pin when 3.7 is dropped
-    "types-requests",
-    "types-urllib3",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -20,9 +20,6 @@
     "pytest-cov==4.*",
     "pytest-vcr==1.*",
     "pytest==7.*",
-    "twine==5.*;python_version>='3.8'",  # TODO: remove python pin when 3.7 is dropped
-    "types-requests",
-    "types-urllib3",
```
jchen293, line 25: Why was this removed?

jchen293 (approved):

nwithan8 (approved):

setup.py
Justintime50: Described in the PR summary.

PR #354, opened 2024-10-24
feat: add tracking_codes as param for all on TrackerService

Justintime50:
# Description

`tracking_codes` is the preferred param to use for filtering tracking codes on the index endpoint - added here

# Testing

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # CHANGELOG
 
-## Next Release
+## v9.5.0 (2024-10-24)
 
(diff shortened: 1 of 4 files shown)


jchen293 (commented):
Can we have a unit test for this

Justintime50:
> Can we have a unit test for this

I don't feel one is warranted. I've hit the API directly with this param and it works as intended. This parameter isn't specific to this library and the function is already tested. Adding another param to an already working function is effectively a no-op.

nwithan8 (approved):

PR #355, opened 2025-01-07
feat: adds new cc and bank functions for ReferralCustomers

Justintime50:
# Description
- Adds the following functions to assist ReferralCustomers add credit cards and bank accounts:
  - `BetaReferralCustomerService.retrieve_credit_card_client_secret`
  - `BetaReferralCustomerService.retrieve_bank_account_client_secret`
  - `ReferralCustomerService.add_credit_card_from_stripe`
  - `ReferralCustomerService.add_bank_account_from_stripe`

# Testing

- [x] Adds tests for all functions, unfortunately because we don't have Stripe accounts for these test users we can't easily determine a valid assertion so like other functions of their kind, we assert the adjacent associated error to ensure we hit the right endpoint.

# Pull Request Type


easypost/services/beta_referral_customer_service.py
```diff
@@ -36,33 +37,56 @@ def add_payment_method(
...
+            params=params,
+            beta=True,
+        )
+
+        return convert_to_easypost_object(response=response)
+
+    def retrieve_credit_card_client_secret(self) -> Dict[str, Any]:
+        """Retrieves a client secret to use with Stripe when adding a credit card."""
+        response = Requestor(self._client).request(
+            method=RequestMethod.POST,
```
nwithan8, line 75: An empty POST to get something from the API...

jchen293 (dismissed):

easypost/services/beta_referral_customer_service.py
Justintime50: Super cool huh? 😉 We could call this `create_` instead of `retrieve_` which may make more sense? It is indeed a POST

jchen293: I think it depends on how the API docs content is written. If the API docs state "retrieve a client secret," then the current function naming makes more sense.

nwithan8 (approved):

jchen293 (approved):

Justintime50:
Shelving for now due to prioritization.

